PR #44, opened 2015-01-06
Add fix from monolog

Stype:
A low severity issue, but it would be nice to get this into the checks.


monolog/monolog/2014-12-29-1.yaml
```diff
@@ -0,0 +1,8 @@
+title:     Fixed header injection in NativeMailerHandler
```
stof: The title should not contain ``Fixed``. The vulnerability is not the fix itself.

legoktm:
...bump?

PR #49, opened 2015-02-19
added namshi/jose vulnerability

odino:
Hey guys, we just found a vulnerability in one of our libraries and would like it to be pushed here so that more users will realize they need to upgrade :)

cheers!

We will also publish a post soon with a longer explanations so I will change the `link` in another PR, for now its just a link to the commit that fixes the issue.


odino:
thanks man!

PR #50, opened 2015-02-24
add thelia advisory

lunika:


thelia/backoffice-default-template/2015-02-24-1.yaml
```diff
@@ -0,0 +1,8 @@
+title:     XSS injection in backoffice
+link:      http://thelia.net/version-2-1-2-with-security-fix
+cve:       ~
+branches:
+    2.1.x:
+        time:     2015-02-24 17:46:34
+        versions: [>=2.1.0,<2.1.1]
```
stof: should be ``<2.1.2`` given that the fix is in the 2.1.2 release according to the blog post

lunika: true

thanks

PR #52, opened 2015-03-01
Add ZF2015-02

DavidPrevot:
All ZendFramework files name uses their internal ID name, so I went for it.
Please, do tell if ZF2015-02.yaml should be renamed as CVE-2015-0270.yaml.


fabpot:
I think consistency is better for the file name. Anyway, it does not have any consequences, so what you've done looks good to me. Thanks.

Ocramius:
This should also affect `zendframework/zendframework`

PR #54, opened 2015-03-02
Add ZF2015-02 for zendframework/zendframework too

DavidPrevot:
Thanks @Ocramius for the follow up: I forgot `zendframework/zendframework` in #52, this should fix it now.


Ocramius:
@DavidPrevot thanks for listening :-)

PR #55, opened 2015-03-12
Add ZF2015-03

DavidPrevot:


zendframework/zend-validator/ZF2015-03.yaml
```diff
@@ -0,0 +1,17 @@
+title:     Invalid CSRF validation of null or incorrectly formatted token identifiers
+link:      http://framework.zend.com/security/advisory/ZF2015-03
+cve:       CVE-2015-1786
+branches:
+    2.0.x:
+        time:     2015-03-12 13:58:47
+        versions: [>=2.0.0,<2.0.99]
```
xabbuh: Is this right? The blog post states:

> Note: in testing the patch, we discovered that the vulnerability was introduced specifically in the 2.3 series, and thus no patch was necessary against the 2.2 series.

So I would assume that you can remove the entries for `2.0.x`, `2.1.x` and `2.2.x`.

DavidPrevot: > I would assume that you can remove the entries for `2.0.x`, `2.1.x` and
> `2.2.x`.

Right, the documentation states that `branches` is a hash of affected
branches, so I also fixed the previous ZendFramework entries.

Thanks for the review!

Regards

David

zendframework/zendframework/ZF2014-05.yaml
```diff
@@ -2,12 +2,6 @@ title:     Anonymous authentication in ldap_bind() function of PHP, using null b
 link:      http://framework.zend.com/security/advisory/ZF2014-05
 cve:       ~
 branches:
-    2.0.x:
-        time:     2014-09-16 22:00:00
-        versions: [>=2.0.0,<2.0.99]
-    2.1.x:
-        time:     2014-09-16 22:00:00
-        versions: [>=2.1.0,<2.1.99]
```
xabbuh: Looking at the security advisory, Zend Framework before `1.12.9` was also affected. Though I'm not sure if that version will be covered in the repository at all.

DavidPrevot: Right, that is already documented in https://github.com/FriendsOfPHP/security-advisories/blob/master/zendframework/zendframework1/ZF2014-05.yaml.

xabbuh: Ah thanks, didn't see that these are different databases.

zendframework/zendframework/ZF2014-06.yaml
```diff
@@ -2,12 +2,6 @@ title:     SQL injection vector when manually quoting values for sqlsrv extensio
 link:      http://framework.zend.com/security/advisory/ZF2014-06
 cve:       ~
 branches:
-    2.0.x:
-        time:     2014-09-16 22:00:00
-        versions: [>=2.0.0,<2.0.99]
-    2.1.x:
-        time:     2014-09-16 22:00:00
-        versions: [>=2.1.0,<2.1.99]
```
stof: Are 2.0 and 2.1 really not affected by any of these security issues ?

xabbuh: At least, the blog posts only talk about `2.2` and `2.3`.

DavidPrevot: Le 13/03/2015 05:16, Christophe Coevoet a écrit :

> Are 2.0 and 2.1 really not affected by any of these security issues ?

I may indeed have been too quick in this last clean up: at least the
offending code for ZF2014-06 seems actually present in 2.1.8.

Le 13/03/2015 07:27, Christian Flothmann a écrit :

> At least, the blog posts only talk about `2.2` and `2.3`.

Well, AFAICT, there hasn’t been any 2.1 release since April 2013 (nor
any 2.0 releases since Mars 2012), so they were probable EOL, and not
considered upstream anymore when the announce was issued.

I’ll review that last clean up commit ASAP, and will prepare a partial
revert about it, thanks again for the review.

Regards

David

PR #63, opened 2015-04-20
Add Neos and Flow to security database

kdambekalns:
This adds the existing security advisories for typo3/neos and typo3/flow to the security database.


fabpot:
@kdambekalns Great! Thanks for your contributions, that's much appreciated.

PR #64, opened 2015-05-07
Add ZF2015-04/CVE-2015-3154 about zendframework

DavidPrevot:
Maybe @Ocramius could ack this proposal? After a quick look, I believe ZF1 is also partly affected, is a fix also on its way, should we document zendframework/zendframework1 as affected?


Ocramius:
ACK

Don't think ZF1 is affected. If you think so, then please refer to http://framework.zend.com/security/

PR #67, opened 2015-05-11
Nicer error message if dependencies are not installed

cebe:
added a note that running composer install is required.

### DIFF
--- a/validator.php
+++ b/validator.php
@@ -2,7 +2,11 @@
 
 // validates that all security advisories are valid
 
-require __DIR__.'/vendor/autoload.php';
+if (!is_file($autoloader = __DIR__.'/vendor/autoload.php')) {
+	echo "Dependencies are not installed, please run `composer install` first!\n";
+	exit(1);
+}
+require $autoloader;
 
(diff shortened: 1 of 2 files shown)


validator.php
```diff
@@ -2,7 +2,11 @@
 
 // validates that all security advisories are valid
 
-require __DIR__.'/vendor/autoload.php';
+if (!is_file($autoloader = __DIR__.'/vendor/autoload.php')) {
```
Ocramius: This... is kind of a normal thing in a composer project.

Feels like you're fighting the wrong error to me, or completely misunderstood what kind of project we're talking about...

cebe: yeah but this repo is for any kind of php user who maybe is not aware of composer.

the contrib readme states the following:

> Make sure your file validates by running php validator.php from the root of this project.

It does not mention composer at all. It should be either there in the readme or here. Both would be maximum friendly to newbies.

stof: Please also add it in the readme

cebe: done.

fabpot:
Thank you @cebe.

PR #68, opened 2015-05-15
Add the advisory for the pusher-php-server vulnerability

stof:
The announcement was not done on a public URL but by sending a mail to their customers actually.
I used the URL of the gist describing the way to apply the same validation for people writing their own implementations.
@leggetter @WillSewell is there a better URL to be used in the advisory ?

### DIFF
--- a/pusher/pusher-php-server/2015-05-13.yaml
+++ b/pusher/pusher-php-server/2015-05-13.yaml
@@ -1,0 +1,8 @@
+title:     Exploit in the private channel authentication
+link:      https://blog.pusher.com/update-on-security/
+cve:       ~
+branches:
+    master:
+        time:     2015-05-13 10:53:19
+        versions: [<2.2.1]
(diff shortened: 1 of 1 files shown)


leggetter:
@stof there's a blog post pending. It should be published in a couple of hours MAX. I'll comment back here with the link once it's posted.

stof:
great.

WillSewell:
There is also a blog post by Egor who found the vulnerability which goes into more detail: http://sakurity.com/blog/2015/05/08/pusher.html

We wanted to notify our customers first so that they had some time to update their libraries before it became public.

leggetter:
Blog post now published here https://blog.pusher.com/update-on-security/

stof:
the PR is now updated with the URL of the blog post

stof:
@leggetter on a side note, your blog post throws warning about mixed content on the HTTPS page

PR #72, opened 2015-05-28
[Symfony] add affected 2.7 versions for CVE-2015-4050

xabbuh:
Thanks to @GrahamCampbell for noticing this.

### DIFF
--- a/symfony/http-kernel/CVE-2015-4050.yaml
+++ b/symfony/http-kernel/CVE-2015-4050.yaml
@@ -14,4 +14,7 @@ branches:
     2.6.x:
         time:     2015-05-26 23:55:51
         versions: [>=2.6.0,<2.6.8]
+    2.7.x:
+        time:     2015-05-26 23:55:51
+        versions: [>=2.7.0-dev,<2.7.0]
 reference: composer://symfony/http-kernel
--- a/symfony/symfony/CVE-2015-4050.yaml
+++ b/symfony/symfony/CVE-2015-4050.yaml
@@ -14,4 +14,7 @@ branches:
     2.6.x:
         time:     2015-05-26 23:55:51
         versions: [>=2.6.0,<2.6.8]
+    2.7.x:
+        time:     2015-05-26 23:55:51
+        versions: [>=2.7.0-dev,<2.7.0]
 reference: composer://symfony/symfony


symfony/http-kernel/CVE-2015-4050.yaml
```diff
@@ -14,4 +14,7 @@ branches:
     2.6.x:
         time:     2015-05-26 23:55:51
         versions: [>=2.6.0,<2.6.8]
+    2.7.x:
+        time:     2015-05-26 23:55:51
+        versions: [>=2.7.0-dev,<2.7.0]
```
cebe: doesn't `2.7.0-dev` refer to the latest dev release? This may be greater than the upper bound.
Should be something like `>2.6.99999` or is there a better syntax?

stof: @cebe no. You are confusing ``2.7.0-dev`` (dev snapshot tag, so before alphas) and ``2.7.x-dev`` which is the branch version.

And btw, this is the reason why I don't like the possibility to omit the ``.x`` part when specifying branch aliases (having composer add it silently) as it means ``2.7-dev`` has a different meaning only in this context

cebe: I see, thanks for clarification. I did not know about this syntax, and I did not find it in the documentation, at least [here](https://getcomposer.org/doc/01-basic-usage.md#package-versions) it is not covered.

There is still a problem with the contraint. This matches:
http://semver.mwl.be/#?package=symfony%2Fsymfony&version=%3E%3D2.7.0-dev&minimum-stability=dev
but this does not match anymore:
http://semver.mwl.be/#?package=symfony%2Fsymfony&version=%3E%3D2.7.0-dev,%3C2.7.0&minimum-stability=dev

stof: well, it looks like semver.mwl.be does not consider than a beta version is below the corresponding stable version: http://semver.mwl.be/#?package=symfony%2Fsymfony&version=%3C2.7.0&minimum-stability=dev does not match the 2.7.0 beta releases while it should.

cebe: I have created a bug report for composer: https://github.com/composer/composer/issues/4080

fabpot:
I did not mention the 2.7 beta versions in the blog post on purpose. Non stable versions are not covered by security fixes and should not be used in production anyway.

PR #73, opened 2015-05-28
Add recent CakePHP security issues.

markstory:
I'm not sure if I did this right. I'm more than happy to massage these into a more acceptable format if I've made a mistake.


cakephp/cakephp/2015-05-28.yaml
```diff
@@ -0,0 +1,11 @@
+title:     Denial of Service attack through XML payloads
+link:      http://bakery.cakephp.org/2015/05/28/cakephp_2_6_6_and_3_0_6_released.html
+cve:       ~
+branches:
+    3.0.x:
+        time:     2015-05-28 11:30:00
+        versions: [>=3.0.0,<3.0.6]
+    2.x.x:
```
fabpot: `2.x.x` is not a valid notation. You should list all possible values here (`2.0.x`, `2.1.x`, ...).

markstory: Fixed up I think.

markstory:
Thanks @fabpot

PR #77, opened 2015-08-03
Adds ZF2015-06 / CVE-2015-5161

weierophinney:
- Affects zendframework/zendframework
- Affects zendframework/zendframework1
- Affects zendframework/zendxml

See http://framework.zend.com/security/advisory/ZF2015-06 for details.

### DIFF
--- a/zendframework/zendframework/ZF2015-06.yaml
+++ b/zendframework/zendframework/ZF2015-06.yaml
@@ -1,0 +1,23 @@
+title:     XXE/XEE vector when using ZendXml on multibyte payloads
+link:      http://framework.zend.com/security/advisory/ZF2015-06
+cve:       CVE-2015-5161
+branches:
+    2.0.x:
+        time:     2015-08-03 15:13:58
+        versions: [>=2.0.0,<2.0.99]
+    2.1.x:
+        time:     2015-08-03 15:13:58
+        versions: [>=2.1.0,<2.1.99]
+    2.2.x:
+        time:     2015-08-03 15:13:58
+        versions: [>=2.2.0,<2.2.99]
+    2.3.x:
+        time:     2015-08-03 15:13:58
+        versions: [>=2.3.0,<2.3.8]
+    2.4.x:
+        time:     2015-08-03 15:13:58
+        versions: [>=2.4.0,<2.4.6]
+    2.5.x:
+        time:     2015-08-03 15:13:58
+        versions: [>=2.5.0,<2.5.1]
+reference: composer://zendframework/zendframework
--- a/zendframework/zendxml/ZF2015-06.yaml
+++ b/zendframework/zendxml/ZF2015-06.yaml
@@ -1,0 +1,8 @@
+title:     XXE/XEE vector when using ZendXml on multibyte payloads
+link:      http://framework.zend.com/security/advisory/ZF2015-06
+cve:       CVE-2015-5161
+branches:
+    1.0.x:
+        time:     2015-08-03 15:13:58
+        versions: [>=1.0.0,<1.0.1]
+reference: composer://zendframework/zendxml
--- a/zendframework/zendframework1/ZF2015-06.yaml
+++ b/zendframework/zendframework1/ZF2015-06.yaml
@@ -1,0 +1,8 @@
+title:     XXE/XEE vector when using ZendXml on multibyte payloads
+link:      http://framework.zend.com/security/advisory/ZF2015-06
+cve:       CVE-2015-5161
+branches:
+    1.12.x:
+        time:     2015-08-03 15:13:58
+        versions: [>=1.12.0,<1.12.14]
+reference: composer://zendframework/zendframework1


zendframework/zendxml/ZF2015-06.yaml
```diff
@@ -0,0 +1,8 @@
+title:     XXE/XEE vector when using ZendXml on multibyte payloads
+link:      http://framework.zend.com/security/advisory/ZF2015-06
+cve:       CVE-2015-5161
+branches:
+    1.0.x:
+        time:     2015-08-03 15:13:58
+        versions: [>=1.0.0,<1.0.1]
+reference: composer://zendframework/zendframework1
```
Ocramius: Reference seems to point to the wrong package

cebe: shouldn't this be `zendframework/zendxml` ?

zendframework/zendframework/ZF2015-06.yaml
```diff
@@ -0,0 +1,20 @@
...
+link:      http://framework.zend.com/security/advisory/ZF2015-06
+cve:       CVE-2015-5161
+branches:
+    2.0.x:
+        time:     2015-08-03 15:13:58
+        versions: [>=2.0.0,<2.0.99]
+    2.1.x:
+        time:     2015-08-03 15:13:58
+        versions: [>=2.1.0,<2.1.99]
+    2.3.x:
```
cebe: 2.2.x is not affected?

weierophinney: oops - yes, it is. Will update.

Ocramius:
:+1:

cebe:
:+1:

zendframework/zendframework/ZF2015-06.yaml
```diff
@@ -0,0 +1,23 @@
...
+        versions: [>=2.2.0,<2.2.99]
+    2.3.x:
+        time:     2015-08-03 15:13:58
+        versions: [>=2.3.0,<2.3.8]
+    2.4.x:
+        time:     2015-08-03 15:13:58
+        versions: [>=2.4.0,<2.4.6]
+    2.5.x:
+        time:     2015-08-03 15:13:58
+        versions: [>=2.5.0,<2.5.1]
```
stof: is ZF 2.5.x really affected ? It does not contain the affected code. It *depends* on zendxml. This means that using 2.5.0 is fine as long as the newer zendxml is used

weierophinney: If you install at the lowest dependency values, you'll get ZendXml 1.0.0, which *is* affected. For this reason, we issued a 2.5.2 release that pins to `^1.0.1`.

stof: but in this case, the security-checker will already warn you about the security issue in zendxml. There is no need to warn about the zf2 package if it is not the one containing the vulnerability

Ocramius: True, but this makes it more explicit, imo

stof: @Ocramius but it reports vulnerabilities in non-vulnerable software. and it would report the vulnerability twice in vulnerable software (because it would still be reported in ZendXml)

Ocramius: True that. I guess you're right here :)

weierophinney: @stof While *one* use case for this repository is to provide machine-readable information that can be then consumed by automated tooling, another *stated purpose* in the readme is to allow *browsing existing vulnerabilities*.

Zend Framework provides distribution packages for ZF2 that pull in both the framework *and* ZendXml. For users of those download packages, if they have an earlier version *they are vulnerable*.

Please remember that *not everyone uses composer*. This is particularly true in enterprises (I hear it all the time particularly from enterprises using IBM i Series). As such, I'd argue that being more explicit here about where the vulnerabilities exist is better than not being explicit enough, and giving somebody a false sense of security.

stof: @weierophinney if you go this way, you should also reference all projects based on ZF2 and distributing packaged code.
These files are referencing vulnerabilities affecting composer packages (see the reference on the next line). The ``zendframework/zendframework`` composer package is not vulnerable. The distribution archive on the ZF website is. But this distribution archive contains multiple packages

weierophinney: @stof Then I respectfully ask that you change the README to reflect that the repository is only for:

- Machine readable vulnerability information, that
- is only usable with packages installable via composer
- and specifically referenced in packagist.org.

Right now, the repository README file indicates the following:

> The PHP Security Advisories Database references known security vulnerabilities in various PHP projects and libraries. This database must not serve as the primary source of information for security issues, it is not authoritative for any referenced software, but it allows to centralize information for convenience and easy consumption.

It specifically indicates that it is a *database*, and indicates that you can determine if a version is vulnerable in *multiple* ways:

- Via an online, browseable database
- A CLI tool
- An API

Nowhere does it indicate that the vulnerabilities listed are limited in the fashion I suggested above. As such, I have listed the vulnerabilities *as we have reported them and mitigated them in our project.* 

I'm leaving the PR as-is unless changes to the README to limit the scope of this repository are made.

stof: But references are only supporting composer references currently. There is no other reference type supported currently (even though the way they are stored makes it possible to support other reference types in the future).

the issue is that this advisory is saying that the ``zendframework/zendframework`` *composer package* is vulnerable in its 2.5.0 version. and this information is wrong.

It is possible to add advisories for packages not on Packagist btw (though it looks weird).

weierophinney: > It is possible to add advisories for packages not on Packagist btw (though it looks weird).

Can you provide examples for me, then?

stof: well, the advisory database just needs to know the package name. It does not care whether it is on Packagist or no. You can still submit an advisory for packages other packages by using their name (but if it is a private package, it looks weird to submit it to a public database)

fabpot: I think it's up to each project to determine how they want to advertize vulnerabilities. I agree with @stof, but ultimately, it's @weierophinney call. So, I'm going to merge it as is.

PR #78, opened 2015-08-04
Added laravel socialite vulnerabilities

GrahamCampbell:
The first vulnerability was in the fact that the generated state wasn't actually random and the user could work it out from their session.

The second vulnerability prevents two things. The first part of this fix prevents guessing the state through multiple attempts, and the second prevents an attacker from gaining access by not sending a session cookie and then sending no state as a query parameter - in this condition, you could bypass state checking all together because the check to check the state wasn't empty was missing so emtpy === empty was passing...

### DIFF
--- a/laravel/socialite/2015-07-23.yaml
+++ b/laravel/socialite/2015-07-23.yaml
@@ -1,0 +1,11 @@
+title:     Insecure state generation
+link:      https://github.com/laravel/socialite/pull/91
+cve:       ~
+branches:
+    1.0.x:
+        time:     2015-07-23 13:53:00
+        versions: [>=1.0.0,<1.0.99]
+    2.0.x:
+        time:     2015-07-23 13:53:00
+        versions: [>=2.0.0,<2.0.9]
+reference: composer://laravel/socialite
--- a/laravel/socialite/2015-08-03.yaml
+++ b/laravel/socialite/2015-08-03.yaml
@@ -1,0 +1,11 @@
+title:     State guessing vulnerability
+link:      https://github.com/laravel/socialite/pull/93
+cve:       ~
+branches:
+    1.0.x:
+        time:     2015-08-03 12:55:00
+        versions: [>=1.0.0,<1.0.99]
+    2.0.x:
+        time:     2015-08-03 12:55:00
+        versions: [>=2.0.0,<2.0.10]
+reference: composer://laravel/socialite


joecohens:
:+1:

laravel/socialite/2015-07-23.yaml
```diff
@@ -0,0 +1,11 @@
+title:     Insecure state generation
+link:      https://github.com/laravel/socialite/pull/91
```
stof: it would be good to have a link to something describing the issue in more details. The PR doing the patch does not contain any information (what is the severity of the issue, which cases are affected, ...)

stof: btw, the explanation given in https://github.com/laravel/socialite/pull/93#discussion_r36254365 is precisely related to the fact that the actual issues and the action taken are not described

GrahamCampbell: We didn't document this security fix, sorry. That pull request is the only thing we have in writing.

GrahamCampbell: That comment on the PR provides sufficient documentation now anyway.

stof: @GrahamCampbell what about documenting them then ? It seems a good idea

GrahamCampbell: There are no plans to document it sorry. I'd just like to have it documented here.

cebe: > There are no plans to document it sorry.

just having an information about the fact that there is some kind of security issue but no information about the issue details is not very helpful.
If you find a security issue in your framework or library you should have a way to inform your users and let them know how severe the issue is and how they can protect themselfs. This information is important so that users can decide on whether they are vulnerable or not and in which way.
It is also recommended to reserve a CVE number for relevant security issues: https://cve.mitre.org/cve/request_id.html Assignment is as easy as asking for one by writing to cve-assign@mitre.org. CVE provides a general way of dealing with security issues and keep track of the information on them.

stof: @GrahamCampbell the issue is that this advisory database will reference the security issue, but it will not document it. There will be no explanation about the issue (this is precisely what the link given in the advisory is meant to cover)

GrahamCampbell: I'll talk to Taylor about this. Ultimately, it's up to him, not me.

Ocramius: Note that just providing some details in the PR (now that it is merged) is ok/sufficient.

GrahamCampbell: Yeh, that's what we thought @Ocramius. I'm still going to let Taylor know about this feedback though. :)

PR #79, opened 2015-08-07
Add new CakePHP issue.

markstory:
We had a new issue in the 2.x release series that warrants more visibility on.


markstory:
Thanks :+1:

PR #82, opened 2015-08-12
Switch to the docker-based infrastructure on Travis

stof:
The new infrastructure has a faster boot time, and more capacity, so it is better to use it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,12 @@
 language: php
 
+sudo: false
+
+cache:
+    - $HOME/.composer/cache/files
+
 install:
-    - travis_retry composer install --no-interaction --prefer-source
+    - travis_retry composer install
 
 script:
     - php validator.php


GrahamCampbell:
It actually has less capacity, but fewer people use it. :P

.travis.yml
```diff
@@ -1,7 +1,12 @@
 language: php
 
+sudo: false
+
+cache:
+    - $HOME/.composer/cache/files
+
 install:
-    - travis_retry composer install --no-interaction --prefer-source
+    - travis_retry composer install
```
GrahamCampbell: You need to keep the no interaction.

stof: nope. Travis forces it through environment variable now (since a long time)

stof:
> It actually has less capacity, but fewer people use it. :P

they are scaling the capacity much more easily AFAICT, meaning it does not block when there is more demand: it scales.

.travis.yml
GrahamCampbell: Ahh, great. I never knew. :)

stof: this was done in https://github.com/travis-ci/travis-cookbooks/commit/ce17f7a2b61cf026e279ec6cc84913b79b908e71

fabpot:
Thank you @stof.

PR #83, opened 2015-08-31
Advisory - doctrine - CVE-2015-5723

Ocramius:
See http://www.doctrine-project.org/2015/08/31/security_misconfiguration_vulnerability_in_various_doctrine_projects.html - CVE-2015-5723


doctrine/annotations/2015-08-31.yaml
```diff
@@ -0,0 +1,14 @@
+title:     Security Misconfiguration Vulnerability in various Doctrine projects
+link:      http://www.doctrine-project.org/2015/08/31/security_misconfiguration_vulnerability_in_various_doctrine_projects.html
+cve:       CVE-2015-5723
+branches:
+    1.0.x:
```
Ocramius: Do non-existing branches make any difference? I couldn't find any way to match ranges that are not covered by a branch anymore.

doctrine/annotations/2015-08-31.yaml
```diff
@@ -0,0 +1,14 @@
...
+        time:     2015-08-31 14:22:00
```
Ocramius: I'll have to check for the tag date first.

doctrine/annotations/2015-08-31.yaml
stof: no it does not. You could even make a single range ``<1.2.7`` in the 1.2.x branch

Ocramius: Gonna do that now then.

Ocramius: Much better.

Ocramius:
@fabpot thanks!

PR #84, opened 2015-09-01
Correcting YAML syntax for branch names

Ocramius:
Also, this is reason 8400347509437 for YAML being a shitty configuration format.

Fitting quote:

 > The story so far:
 > In the beginning the Universe was created.
 > This has made a lot of people very angry and been widely regarded as a bad move.

### DIFF
--- a/doctrine/common/2015-08-31.yaml
+++ b/doctrine/common/2015-08-31.yaml
@@ -2,10 +2,10 @@ title:     Security Misconfiguration Vulnerability in various Doctrine projects
 link:      http://www.doctrine-project.org/2015/08/31/security_misconfiguration_vulnerability_in_various_doctrine_projects.html
 cve:       CVE-2015-5723
 branches:
-    2.4:
+    "2.4":
         time:     2015-08-31 14:42:59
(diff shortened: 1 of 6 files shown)


GrahamCampbell:
:+1:

stof:
this looks like something which should be added to the validator script to be automated, no ?

Ocramius:
@stof not sure if it can be automated, as it fails in the Yaml component itself. Maybe an addition to the YAML stuff?

stof:
@Ocramius A solution could be to enforce branches not to have numeric names in the validator (the branch is generally set as ``1.x`` or ``2.4.x`` for versionned branches). This would solve it as this would parse as a Yaml string.

What we could do in the Yaml component to avoid loosing float keys is to cast them to string before adding them to the array instead of letting PHP cast it (which will cast to int)

Ocramius:
> What we could do in the Yaml component to avoid loosing float keys

The actual bug, I suppose. Can you report it, if you got time? Currently being kicked out...

Ocramius:
Reported in symfony/symfony#15667

PR #85, opened 2015-09-02
Fix lower bounds for an advisory

stof:
This issue was fixed in each branch, so we need lower bounds on newer branches. Other 1.1.2 will still be considered as impacted.

### DIFF
--- a/namshi/jose/2015-02-19.yaml
+++ b/namshi/jose/2015-02-19.yaml
@@ -7,11 +7,11 @@ branches:
         versions: [<1.1.2]
     "1.2":
         time:     2015-02-19 10:59:35
-        versions: [<1.2.2]
+        versions: [>=1.2.0,<1.2.2]
     "2.0":
         time:     2015-02-19 10:59:35
-        versions: [<2.0.3]
+        versions: [>=2.0.0,<2.0.3]
     "2.1":
         time:     2015-02-19 10:59:35
-        versions: [<2.1.2]
+        versions: [>=2.1.0,<2.1.2]
 reference: composer://namshi/jose
--- a/validator.php
+++ b/validator.php
@@ -84,6 +84,8 @@ foreach ($dir as $file) {
             continue;  // Don't validate branches when not set to avoid notices
         }
 
+        $upperBoundWithoutLowerBound = null;
+
         foreach ($data['branches'] as $name => $branch) {
             if (!preg_match('/^([\d\.\-]+(\.x)?(\-dev)?|master)$/', $name)) {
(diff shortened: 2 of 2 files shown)


stof:
I added a validation rule to avoid such mistakes in the future. Only the first branch is allowed to omit the lower bound now. This is a simplification of the validation though: a proper validation would allow omitting the lower bound for the branch with the lowest max bound actually (and they could allow several ones to omit it if they have the same max bound)

doctrine/doctrine-module/2013-05-16.yaml
```diff
@@ -2,10 +2,10 @@ title:     Authentication Vulnerability - possible attempt to login via zero-val
...
 branches:
-    master:
-        time:     2013-05-16 18:29:36
-        versions: [<=0.7.1]
     0.7.x-dev:
         time:     2013-05-16 18:27:30
         versions: [<=0.7.1]
+    master:
+        time:     2013-05-16 18:29:36
+        versions: [>=0.7.0,<=0.7.1] # the master branch was fixed separately must was not yet released
```
staabm: I guess there is 1 or 2 words too much in this comment. for me it doesnt make sense ;)

Ocramius: Is this entry even needed then?

stof: well, it depends how the SensioLabs security checker handles the validation of affected dev versions. The 0.7.x and master branches have been fixed at different times.

Changing the constraint in this one was actually necessary just because I was lazy about making the validator more complex (as the max bound is the same in both branches, it may be accepted by a more advanced validation implementation)

stof: but yeah, there is indeed an extra word

doctrine/doctrine-module/2013-05-16.yaml
```diff
@@ -2,10 +2,10 @@ title:     Authentication Vulnerability - possible attempt to login via zero-val
 link:      https://github.com/doctrine/DoctrineModule/issues/249
 cve:       ~
 branches:
-    master:
-        time:     2013-05-16 18:29:36
-        versions: [<=0.7.1]
     0.7.x-dev:
         time:     2013-05-16 18:27:30
         versions: [<=0.7.1]
+    master:
```
Ocramius: I still suggest simply removing this: `master` should not be used anyway

stof:
OK, I reworked a bit the validation, and we now accept multiple branches without lower bound in case they have the same upper bound (as this case is not an overlapping). This avoids having to make a change to the doctrine/doctrine-module advisory just to make the simplistic validator happy

PR #103, opened 2015-09-15
SS 2015 020

spekulatius:
Is the branch definition okay like this? This issue should only apply to this particular branch afaik.

```
  versions: [3.1.14-rc1]
```


spekulatius:
I'm unsure about the notation for one particular branch so I'll close this for now.

PR #104, opened 2015-09-15
Advisories for ZF2015-07 and ZF2015-08

weierophinney:


zendframework/zend-cache/ZF2015-07.yaml
```diff
@@ -0,0 +1,11 @@
+title:     Filesystem Permissions Issues in Multiple Components
+link:      http://framework.zend.com/security/advisory/ZF2015-07
+cve:       CVE-2015-5723
+branches:
+    2.4.x:
+        time:     2015-09-15 16:24:04
+        versions: [>=2.4.0,<2.4.8]
```
stof: Are older 2.x versions not affected ? If they are, the lower bound should be changed, even if there is no fixed version in the 2.3 serie

weierophinney: Older versions are affected; however, I'm uncertain how to indicate those.
Do I need a "branch" entry per minor version? How do I indicate all
versions on a minor branch?

stof: you have 2 choices:

- specifying all branches separately (this is what is done in Symfony generally) with constraints like ``>=2.2.0,<2.3.0``
- changing the lower bound of this branch reference to be ``>=2.0,<2.4.8``

The result would be the same in all cases: all 2.2.x versions would be matched by one of the constraint, marking them as affected

PR #109, opened 2015-10-29
Validate Composer packages

thewilkybarkid:
Although I don't use it, I've spotted that a package seems to have an incorrect Composer name, so I've added a (rather naive) check and corrected it (@spekulatius, could you confirm you meant `silverstripe/cms`?).

This assumes that all Composer packages in this database appear on Packagist (which is currently true, but they don't have to - though would they appear in this database if not?).

I've also noticed that all the SilverStripe advisory URLs are 404s (did they ever work?). Should they be checked too?

### DIFF
--- a/silverstripe/cms/SS-2015-008-1.yaml
+++ b/silverstripe/cms/SS-2015-008-1.yaml
@@ -8,4 +8,4 @@ branches:
     3.1.x:
         time:     2015-03-19 16:54:00
         versions: [>=3.1.0,<3.1.11]
-reference: composer://framework/cms
+reference: composer://silverstripe/cms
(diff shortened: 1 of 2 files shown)


spekulatius:
Hello @thewilkybarkid ,

Thanks for your good spotting - you are right. It should be silverstripe/cms. I can't tell why I wrote framework.

The SilverStripe advisory URLs used to work at the time - after the recent rebuild of the websites they aren't working anymore. I'll check if I get direct links for the advisories somewhere.

Regarding Packagist: I would assume that a package doesn't need to be on Packagist. I could for example add the sources directly and refer to packages not listed at Packagist. Yet I would think this is a very unlikely scenario it still a question. @fabpot might know the answer to.

Cheers,
Spek

fabpot:
Packages do not  need to be on Packagist, but it's a must if you want to use https://security.sensiolabs.org/

PR #110, opened 2015-10-29
Fix composer.json

thewilkybarkid:
These are a few commits that fix some issues in the project's `composer.json`.

I've separated them out in case you don't want some of them; I can squash any/all if needed.


composer.json
```diff
@@ -1,9 +1,10 @@
 {
     "name": "sensiolabs/security-advisories",
-    "require": {
+    "description": "Database of known security vulnerabilities in various PHP projects and libraries",
+    "require-dev": {
```
GrahamCampbell: ?

thewilkybarkid: `symfony/yaml` is only needed by the validator.

GrahamCampbell: ah, right :)

stof:
Why is the lock file removed ?

Issue #111, opened 2015-10-31
Invalid Yaml


reiz:
I just used http://www.yamllint.com/ to double check [this](https://raw.githubusercontent.com/FriendsOfPHP/security-advisories/master/doctrine/annotations/2015-08-31.yaml) file here and it says that's it's not valid. I'm getting this error with multiple yaml files from this repository. Any ideas?

reiz:
Just tried another YAML validator and got the same result. See [here](http://yaml-online-parser.appspot.com/?url=https%3A%2F%2Fraw.githubusercontent.com%2FFriendsOfPHP%2Fsecurity-advisories%2Fmaster%2Fdoctrine%2Fannotations%2F2015-08-31.yaml).

PR #112, opened 2015-10-31
Make Yaml valid.

reiz:


xabbuh:
@reiz In #111 you talked about multiple files. I think it would be a good idea to fix them all.

reiz:
@xabbuh Yes. Multiple files are affected. I noticed that because the VersionEye crawler logged out exceptions for some files. In the mean while I wrote a workaround to deal with corrupted Yaml files. I can output a list of the affected files and send you a bigger pull request later this day.

reiz:
@xabbuh Here is the complete list: https://github.com/FriendsOfPHP/security-advisories/pull/113

PR #115, opened 2015-11-11
Update Neos/Flow advisories

kdambekalns:
Since the projects separated earlier this year, the links to the
bulletins are updated to point to the current (and working) ones.


kdambekalns:
Gah, ok. That makes sense, somehow…

PR #116, opened 2015-11-13
Add recent CakePHP issue.

markstory:


markstory:
Thanks :+1:

cakephp/cakephp/2015-11-05.yaml
```diff
@@ -0,0 +1,35 @@
...
+        versions: [>=2.6.0,<2.6.12]
+    2.7.x:
+        time:     2015-11-05 22:08:00
+        versions: [>=2.7.0,<2.7.6]
+    3.0.x:
+        time:     2015-11-05 22:08:00
+        versions: [>=3.0.0,<3.0.15]
+    3.1.x:
+        time:     2015-11-05 22:08:00
+        versions: [>=3.1.0,<3.1.3]
```
kenjis: @markstory According to the link <http://bakery.cakephp.org/2015/11/05/cakephp_3015_314_2612_276_released.html>, CakePHP 3.1.4 was released.
I'm not sure, but the `<3.1.3` is right?

markstory: You're right these ranges are off by one. I'll fix that.

PR #117, opened 2015-11-22
Add XSS in CodeIgniter

kenjis:
This adds the existing vulnerability in CodeIgniter 3.

### DIFF
--- a/codeigniter/framework/2015-10-31-1.yaml
+++ b/codeigniter/framework/2015-10-31-1.yaml
@@ -1,0 +1,8 @@
+title: XSS attack vector in Security Library method xss_clean()
+link: http://www.codeigniter.com/user_guide/changelog.html#version-3-0-3
+cve: ~
+branches:
+    master:
+        time:     2015-10-27 12:30:18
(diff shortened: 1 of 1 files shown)


codeigniter/framework/2015-10-31-1.yaml
```diff
@@ -0,0 +1,8 @@
+title: XSS attack vector in Security Library method xss_clean()
+link: http://www.codeigniter.com/user_guide/changelog.html#version-3-0-3
+cve: ~
+branches:
+    master:
+        time:     2015-10-27 12:30:18
+        versions: [<3.0.3]
```
xabbuh: Are you sure that `1.x` and `2.x` versions are affected by this too?

kenjis: We can't install `1.x` and `2.x` via Composer.

To be accurate, `2.2.6` is not affected by it. http://www.codeigniter.com/userguide2/changelog.html

I don't know about `1.x`, but probably affected.

And `1.x` and `2.x` reached EOL.

Do I need to add the lower version?

PR #121, opened 2015-11-24
Add ZF2015-10

paragonie-scott:


Ocramius:
This should include zf >2.0.0,<2.4.9 as per http://framework.zend.com/blog/zend-framework-1-12-17-and-2-4-9-released.html

zendframework/zend-crypt/2015-11-23.yaml
```diff
@@ -0,0 +1,11 @@
+title:      Potential Information Disclosure in Zend\Crypt\PublicKey\Rsa\PublicKey
+link:       http://framework.zend.com/security/advisory/ZF2015-10
+cve:        CVE-2015-7503
+branches:
+    2.4.x:
+        time:     2015-11-23 14:30:00
+        versions: [<2.4.9]
+    2.5.x:
+        time:     2015-11-23 14:30:00
+        versions: [<2.5.2]
```
stof: this needs a lower bound, otherwise 2.4.10 will also be mark as impacted

stof: btw, this is why tests are failing

paragonie-scott:
Updated and squashed the commits. :)

Ocramius:
Still needs to cover `zendframework/zendframework` as well tho :-\

paragonie-scott:
How would you recommend doing that?

zendframework/zend-crypt/2015-11-23.yaml
```diff
@@ -0,0 +1,11 @@
+title:      Potential Information Disclosure in Zend\Crypt\PublicKey\Rsa\PublicKey
+link:       http://framework.zend.com/security/advisory/ZF2015-10
+cve:        CVE-2015-7503
+branches:
+    2.4.x:
+        time:     2015-11-23 14:30:00
+        versions: [>=2.4.0,<2.4.9]
```
stof: according to @Ocramius's comment, the lower bound should be 2.0.0 here

stof:
@paragonie-scott create a second advisory for the ``zendframework/zendframework`` package (with the same content except for the composer reference)

zendframework/zendframework/ZF2015-10.yaml
```diff
@@ -0,0 +1,11 @@
+title:      Potential Information Disclosure in Zend\Crypt\PublicKey\Rsa\PublicKey
+link:       http://framework.zend.com/security/advisory/ZF2015-10
+cve:        CVE-2015-7503
+branches:
+    2.4.x:
+        time:     2015-11-23 14:30:00
+        versions: [>=2.0.0,<2.4.9]
+    2.5.x:
```
Ocramius: this key should be dropped (after 2.5.0, all packages are independent. Sadly, our LTS is 2.4.x)

Ocramius:
:+1:

PR #122, opened 2015-11-24
Adding Flow-SA-2015-001 and Neos-SA-2015-002

kdambekalns:


stof:
shouldn't we have 2 advisories, one for the arbitrary file upload and another one for the XEE ?

kdambekalns:
We have released one Flow advisory, and fixed both issues in the same release. To me it seems the entries here should be consistent with the advisories… No?

stof:
but the affected versions are not the same for both issues

kdambekalns:
Still seemed to make sense to have a "combined advisory" from our POV. But you have a point there, and with the next advisory (hopefully will take a while!) we'll keep that in mind.

As of today, as I see it, this database reflects the advisories released, not the other way around… So I'd be happy if this could be merged, unless the API and/or CLI checker have an issue with the current PR. But looking at the data, it should detect that there is an issue for the affected versions and point to the advisory. So it should be fine, no?

stof:
@kdambekalns this was mostly a suggestion. It is not preventing the merge IMO (but I'm not a merger here)

kdambekalns:
Ok, I see. Anyway, thanks!

PR #124, opened 2015-11-24
[Symfony] mark unmaintained versions as affected

xabbuh:
Adds constraints for the unsupported `2.4.x` and `2.5.x` Symfony versions so that they are reported as being affected by CVE-2015-8124 and CVE-2015-8125 too.


fabpot:
Thank you @xabbuh.

Issue #107, opened 2015-09-27
extend the reference logic


Flyingmana:
Hi,
I want to start something similar for the Magento ecosystem. Now we have here not such a composer centric ecosystem which results in projects available over different sources. Some of them even with multiple composer Identifiers.

As other smaller ecosystems may have similar problems, I would like to solve it here to keep a central standard for the format.

My ideas would be to introduce an optional Alias argument, or allowing of multiple reference entries.

How do you see this?

fabpot:
Keeping a central repository (for now) seems like a good idea. Having aliases is also possible. Can you paste here a YAML document that would use such aliases?

Flyingmana:
```yaml
title:     Blind SQL Injection flaw
link:      http://appcheck-ng.com/critical-security-flaw-patched-in-magento-blog-extension-cve-2015-3428/
cve:       CVE-2015-3428
branches:
    1.x:
        time:     2015-05-27 12:00:00
        versions: [<1.3]
reference: composer://connect20/aw_blog
aliases:
    magento://AW/Blog
```

fabpot:
Sounds good to me. I suppose it should be:

```yaml
aliases:
    - magento://AW/Blog
```

Flyingmana:
yeah, Iam not used to yaml, your right

syzer:
:+1:

Ocramius:
@Flyingmana few things:

 - did you consider poking the folks working on [PSR-9 and PSR-10](http://blog.phpdeveloper.org/2015/05/22/php-security-psr-9psr-10/)?
 - how is this alias actually "used" outside the composer ecosystem? This tool currently checks `composer.lock`: how do you envision it working for the magento installer stuff?

reiz:
@Ocramius VersionEye is already crawling the magento-security-advisories every 5 minutes. If your composer.lock file is monitored by VersionEye you will receive automatically an email notification if one of your dependencies is affected.

fabpot:
Closing as there is no activity here.

PR #126, opened 2015-11-29
[TASK] Add (almost) all security bulletins for TYPO3 community extensions

mostwanted1976:
This commit adds almost (!) all security bulletins for every community extension
of the CMS "TYPO3" hosted on TYPO3's "TYPO3 extension repository" (TER).

As these bulletins need some more manual work all missing security bulletins
will be pushed in the next days if this commit is accepted.

'php validator.php' notes that all included packages are 'Invalid composer
package'. As far as I can see this is caused by the fact that TER hosted
TYPO3 extensions are not hosted on packagist.org but on TYPO3's own composer
repository at https://composer.typo3.org/. I hope it is possible to use this
security advisories anyway.

Some information about that can be found at composer.typo3.org.


mostwanted1976:
I am not sure if it is right that I only used branch "master" here.

mostwanted1976:
```
'php validator.php' notes that all included packages are 'Invalid composer
package'. As far as I can see this is caused by the fact that TER hosted
TYPO3 extensions are not hosted on packagist.org but on TYPO3's own composer
repository at https://composer.typo3.org/. I hope it is possible to use this
security advisories anyway.

Some information about that can be found at composer.typo3.org.
```

xabbuh:
@mostwanted1976 looks like that problem is somewhat related to #107 (though here we still have Composer packagist but not hosted on packagist.org)

typo3-ter/cal/2009-12-01.yaml
```diff
@@ -0,0 +1,8 @@
+title:     "TYPO3-SA-2009-019: Blind SQL Injection vulnerability in extension Calendar Base (cal)"
+link:      http://typo3.org/teams/security/security-bulletins/typo3-extensions/typo3-sa-2009-019/
+cve:
+branches:
+    master:
+        time:     2009-12-01 12:00
+        versions: [<=1.2.0]
```
stof: we generally write them as ``<1.2.1`` (where ``1.2.1`` is the release containing the fix)

typo3-ter/cal/2010-03-02.yaml
```diff
@@ -0,0 +1,8 @@
+title:     "TYPO3-SA-2010-005: Blind SQL Injection vulnerability in extension Calendar Base (cal)"
+link:      http://typo3.org/teams/security/security-bulletins/typo3-extensions/typo3-sa-2010-005/
+cve:
```
stof: missing ``~``

stof:
@mostwanted1976 see https://github.com/FriendsOfPHP/security-advisories/issues/127 for my proposal about packages on custom composer repositories

typo3-ter/cal/2009-12-01.yaml
mostwanted1976: The version number containing the fix is not noted in the of the TYPO3 project's security bulletins. It would be really a lot of work to identify those release numbers. Therefore I would be happy if ``<=`` would also do the job.

typo3-ter/cal/2010-03-02.yaml
mostwanted1976: Where? For every missing CVE?

stof: ah no sorry. The empty node is a valid representation of ``null`` in Yaml (in case it is possibl to use it). but including it would be safer though

typo3-ter/cal/2009-12-01.yaml
stof: well, it should do it

typo3-ter/cal/2010-03-02.yaml
xabbuh: Imo adding `~` would make clear that the CVE link was left empty on purpose and was not simply forgotten. So I agree with @stof on this.

mostwanted1976: Thx. Done.

typo3-ter/cal/2009-12-01.yaml
mostwanted1976: great!

fabpot:
Now that we support external Composer repository, anyone willing to finish this one?

Issue #127, opened 2015-11-30
Support composer packages not registered on Packagist


stof:
Since #109, we are validating the composer reference for mistakes, by checking the package against packagist. This makes it impossible to register a package available on custom composer repositories (see https://github.com/FriendsOfPHP/security-advisories/pull/126 for typo3 extensions, but it could also affect other communities which only start their migration to a composer ecosystem and keep things outside Packagist).

I suggest to add an optional `composer-repository` setting in the advisory, which would allow to validate the package against a different repository (omitting it would validate against Packagist). What do you think ?

xabbuh:
:+1: sounds like a good idea

fabpot:
:+1:

mostwanted1976:
Any news here? My pull request https://github.com/FriendsOfPHP/security-advisories/pull/126 depends (will depend) on this feature.

PR #128, opened 2015-11-30
Check any Composer repository

thewilkybarkid:
This is an attempt to implement #127 (and would allow #126 to pass). I haven't used the `ComposerRepository` class before, but couldn't see a simple way to consistently find out if a repository knew a particular package name, so it's doing a search then cycling through the results.


mostwanted1976:
I am not sure if e.g. Satis supports such a unique URL for each package.

thewilkybarkid:
I've allowed `composer-repository` to be `false` to cater for #165.

fabpot:
Thank you @thewilkybarkid.

PR #132, opened 2015-12-12
Hotfix: yaml syntax error fixes

Ocramius:
Fixes minor issues inadvertently introduced by #131

This goes on my list of "why NO to YAML" btw :-D


fabpot:
Thanks, good catches

Ocramius:
@fabpot thanks for the quick merge ;-)

xabbuh:
@Ocramius Thanks for the fix! We should have a look why the tests for #131 even passed with these issues.

Ocramius:
Ah, sorry for not adding tests: I caught this upstream while working on
rave/security-advisories (own test suite depends on this package's
contents)

xabbuh:
I thought so, but it's weird that the included validation script did not spot them.

Ocramius:
@xabbuh that's because it doesn't really check anything. The block at https://github.com/FriendsOfPHP/security-advisories/blob/7f01ec80c593ac9c234184a4fadb08bacd65a937/validator.php#L117-L125 (below):

```php
                foreach ($branch['versions'] as $version) {
                    if ('<' === substr($version, 0, 1)) {
                        $upperBound = $version;
                        continue;
                    }
                    if ('>' === substr($version, 0, 1)) {
                        $hasMin = true;
                    }
                }
```

Should be something like a regex that matches:

```php
foreach ($branch['versions'] as $version) {
    if (! preg_match('/^(\\<|\\>)(=){0,1}(\d+)(\.\d+)*$/', $version)) {
       throw new MichaelBay::explosions();
    }
}
```

I'll send a patch right away.

xabbuh:
@Ocramius I would have expected the YAML parser to throw a `ParseException` before. If I didn't miss anything the file didn't even contain valid YAML before.

xabbuh:
Ah never mind, the single quote is treated as part of the unquoted string. :(

Ocramius:
@xabbuh isn't that beautiful? (hint: it isn't ;-) )

PR #133, opened 2015-12-12
#132 - Improving validator to display violations in constraint strings

Ocramius:
This test adds validation required for the changes in #132

Note: this restricts version constraints accepted in this repository.
Note: will fail on the PR branch (on purpose) with an output like following:

```
codeigniter/framework/2015-10-31-1.yaml
    Version constraint "<3.0.3'" is not acceptable.
doctrine/doctrine-bundle/2015-08-31.yaml
    Version constraint "<1.5.2'" is not acceptable.
thelia/thelia/2015-04-13-1.yaml
    Version constraint ">=2.1.0-beta1" is not acceptable.
zendframework/zendframework1/ZF2010-03.yaml
    Version constraint "<1.8.5'" is not acceptable.
```

Once merged into `master`, it shouldn't report any failures.

### DIFF
--- a/validator.php
+++ b/validator.php
@@ -33,6 +33,11 @@ $advisoryFilter = function (SplFileInfo $file) {
     return true; // any other file gets checks and any other folder gets iterated
 };
 
+$isAcceptableVersionConstraint = function ($versionString) {
+    return (bool) preg_match('/^(\\<|\\>)(=){0,1}(\d+)(\.\d+)*$/', $versionString);
+};
+
+/* @var $dir \SplFileInfo[] */
 $dir = new \RecursiveIteratorIterator(new RecursiveCallbackFilterIterator(new \RecursiveDirectoryIterator(__DIR__), $advisoryFilter));
 foreach ($dir as $file) {
     if (!$file->isFile()) {
@@ -115,6 +120,10 @@ foreach ($dir as $file) {
                 $upperBound = null;
                 $hasMin = false;
                 foreach ($branch['versions'] as $version) {
+                    if (! $isAcceptableVersionConstraint($version)) {
+                        $messages[$path][] = sprintf('Version constraint "%s" is not in an acceptable format.', $version);
+                    }
+
                     if ('<' === substr($version, 0, 1)) {
                         $upperBound = $version;
                         continue;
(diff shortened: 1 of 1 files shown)


xabbuh:
Well, it doesn't fail as Travis runs the build on the merge commit. :)

validator.php
```diff
@@ -33,6 +33,11 @@
     return true; // any other file gets checks and any other folder gets iterated
 };
 
+$isAcceptableVersionConstraint = function ($versionString) {
+    return (bool) preg_match('/^(\\<|\\>)(=){0,1}(\d+)(\.\d+)*(-(alpha|beta|rc)\d+){0,1}$/', $versionString);
```
xabbuh: Strictly seen this would also allow something like `>=09` but that's probably negligible (we could of course also change the regex to something like `/^(\\<|\\>)(=){0,1}([1-9]\d*)(\.\d+)*(-(alpha|beta|rc)\d+){0,1}$/`).

Ocramius: Fixing

validator.php
```diff
@@ -33,6 +33,11 @@
     return true; // any other file gets checks and any other folder gets iterated
 };
 
+$isAcceptableVersionConstraint = function ($versionString) {
+    return (bool) preg_match('/^(\\<|\\>)(=){0,1}([1-9]\d*)(\.[1-9]\d*)*(-(alpha|beta|rc)[1-9]\d*){0,1}$/', $versionString);
```
xabbuh: Looks like we now fail to handle versions just having `0` parts.

This should work:

```
/^(\\<|\\>)(=){0,1}(([1-9]\d*)|0)(\.([1-9]\d*)|0)*(-(alpha|beta|rc)[1-9]\d*){0,1}$/

Ocramius: Oh, darn =_= Thanks, sorry for the hasty fix before.

Ocramius:
Meh, reverting to `[WIP]` state - I'm too sleepy to understand what I'm doing wrong.

Ocramius:
Zis be readi for merce.

xabbuh:
:+1:

fabpot:
Thanks @Ocramius, that's much appreciated.

PR #134, opened 2015-12-14
Add joomla/session RCE vulnerability

mbabker:
Ref: https://developer.joomla.org/security-centre/637-20151205-session-remote-code-execution-vulnerability.html

### DIFF
--- a/joomla/session/2015-12-14.yaml
+++ b/joomla/session/2015-12-14.yaml
@@ -4,5 +4,5 @@ cve:       ~
 branches:
     master:
         time:     2015-12-14 18:22:25
-        versions: ['<1.3.0']
+        versions: ['<1.3.1']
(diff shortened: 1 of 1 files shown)


joomla/session/2015-12-14.yaml
```diff
@@ -0,0 +1,8 @@
+title:     Remote Code Execution Vulnerability
+link:      https://developer.joomla.org/security-centre/637-20151205-session-remote-code-execution-vulnerability.html
+cve:       ~
+branches:
+    master:
+        time:     2015-12-14 18:22:25
+        versions: ['<1.3.0']
```
xabbuh: `1.3.0` is affected too according to the blog article.

mbabker: Whoops, fixed it

PR #135, opened 2016-01-11
Add new TYPO3 CMS security advisories

wouter90:


typo3/cms/2015-12-15-1.yaml
```diff
@@ -0,0 +1,28 @@
+title:     "Multiple Cross-Site Scripting vulnerabilities in TYPO3 backend"
+link:      https://typo3.org/teams/security/security-bulletins/typo3-core/typo3-core-sa-2015-011/
+branches:
+    6.2.x:
+        time:     2015-12-15 11:38:00
+        versions: ['>=6.2.0', '<6.2.16']
+    7.0.x:
```
Ocramius: You don't really need to repeat all unsupported versions: two entries are sufficient (`>6,<6.2.16`, `>7,<7.6.1`)

Same applies to other files as well, IMO.

PR #147, opened 2016-04-30
Feature/custom directories

redshark1802:
Currently it was not possible to add custom directories to the validator.
This PR fixes this issue by add an argument that allows to supply an array of directories to be scanned.

This is very useful when combined with #126 and #128 for internal use.


stof:
What is the use case for this ? The validator is aimed at catching invalid submissions in pull requests.

fabpot:
Closing as there is no feedback here.

PR #155, opened 2016-07-19
adds simplesamlphp/simplesamlphp 201606-01: Link injection

marcomenzel:
I'm not really sure about the naming of the file.

For the zendframework issues the name given by the advisory naming (e.g. https://github.com/FriendsOfPHP/security-advisories/blob/master/zendframework/zendframework1/ZF2016-02.yaml), so in this case it should be `201606-01` for the simplesamlphp issue. Following the contribute guide it should be `2016-06-08-1`. In case of doctrine only the date of announcement without an increment is used. Any hints?!


fabpot:
File name is really up to you, this should be consistent for all advisories of a given project though.

PR #157, opened 2016-07-19
adding drupal/drupal with some issues

marcomenzel:
Here are the latest two advisories SA-CORE-2016-002, SA-CORE-2016-003 from drupal/drupal included. 
I'm adding only the 8.1.x branch, because 7.x isn't composer supported and 8.0.x didn't receive the fix.

Please have a closer look at it. 
If everything is alright in here, I would add the other ten issues published in SA-CORE-2016-001 which are 8.x-related.

### DIFF
--- a/drupal/drupal/CVE-2016-5385.yaml
+++ b/drupal/drupal/CVE-2016-5385.yaml
@@ -2,10 +2,10 @@ title:     Drupal Core - Highly Critical - Injection - SA-CORE-2016-003
 link:      https://www.drupal.org/SA-CORE-2016-003
 cve:       CVE-2016-5385
 branches:
-    master:
+    8.0.x:
         time:     2016-07-18 16:01:00
-        versions: ['>=8.0','<8.1.7']
+        versions: ['>=8.0','<8.1.0']
     8.1.x:
         time:     2016-07-18 16:01:00
-        versions: ['>=8.1.0', '<8.1.7']
+        versions: ['>=8.1.0','<8.1.7']
 reference: composer://drupal/drupal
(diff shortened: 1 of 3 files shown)


stof:
if drupal 8.0 is affected but without a fix, it should still be included there (either saying ``>= 8.0.0`` in the range of master, or by adding another range saying ``>= 8.0.0, <8.1.0``.)

Note including 8.0 would mean it is safe, while not receiving the patch means that all versions are vulnerable.

drupal/drupal/CVE-2016-5385.yaml
```diff
@@ -0,0 +1,11 @@
+title:     Drupal Core - Highly Critical - Injection - SA-CORE-2016-003
+link:      https://www.drupal.org/SA-CORE-2016-003
+cve:       CVE-2016-5385
+branches:
+    master:
+        time:     2016-07-18 16:01:00
+        versions: ['>=8.0','<8.1.7']
```
fabpot: This should be 8.1.0

drupal/drupal/CVE-2016-5385.yaml
```diff
@@ -0,0 +1,11 @@
+title:     Drupal Core - Highly Critical - Injection - SA-CORE-2016-003
+link:      https://www.drupal.org/SA-CORE-2016-003
+cve:       CVE-2016-5385
+branches:
+    master:
```
fabpot: This should be 8.0.x instead

marcomenzel:
Hm. I think I got it almost.
```
    8.0.x:
        time:     2016-07-18 16:01:00
        versions: ['>=8.0','<8.1.0']
```
It should be 8.0.x because we're talking about the 8.0.x and not about a master branch. And because of the same reason it should be `<8.1.0` instead of `<8,1.7` which will result in non-overlapping version constraints?

GrahamCampbell:
The "branches" thing is poorly named surely? It should have been named "versions" or something?

marcomenzel:
I'm not sure about that. `branches` is mostly right. Usually there are different stable branches for different supported versions. The usage of `master` is some kind of confusing, I don't really get in which cases it must be mentioned.
For example in 
https://github.com/FriendsOfPHP/security-advisories/blob/master/simplesamlphp/simplesamlphp/201606-01.yamlthe master branch is used, but there are stable branches for each minor release.

PR #164, opened 2016-07-21
duplicating drupal/drupal advisories to drupal/core

### LINKED ISSUE #163: Add drupal/core
Could you please add [drupal/core](https://packagist.org/packages/drupal/core) ? It's a subtree split of [drupal/drupal](https://packagist.org/packages/drupal/drupal) so it should be the same.

marcomenzel:
Following the contribution guide "If some affected code is available through different Composer entries (like when you have read-only subtree splits of a main repository), duplicate the information in several files." I just duplicated the advisories. Should fix #163.

### DIFF
--- a/drupal/core/CVE-2016-3162.yaml
+++ b/drupal/core/CVE-2016-3162.yaml
@@ -1,0 +1,8 @@
+title:     File upload access bypass and denial of service
+link:      https://www.drupal.org/SA-CORE-2016-001
+cve:       CVE-2016-3162
+branches:
+    8.0.x:
+        time:     2016-02-15 18:57:00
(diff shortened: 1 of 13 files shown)


michaelcullum:
Could it be worth adding a feature to just support multiple packages that an advisory affects? Otherwise the duplication could become quite messy.

marcomenzel:
Since some vendors `frequently` rename their packages this should be really appreciated.

cebe:
if the content is exactly the same, we could just create symlinks?

marcomenzel:
It is not exactly the same. In each yaml there's a reference line like `reference: composer://drupal/core`. I don't know if this is necessary.

PR #167, opened 2016-09-08
Add SQLi in CodeIgniter

kenjis:
The validator says:
> Invalid branch name "3.0-stable"

But I don't know how to do.

### DIFF
--- a/codeigniter/framework/2016-07-26-1.yaml
+++ b/codeigniter/framework/2016-07-26-1.yaml
@@ -1,0 +1,8 @@
+title: Critical SQL injection bug in the ODBC database driver
+link: http://forum.codeigniter.com/thread-65803.html
+cve: ~
+branches:
+    3.0-stable:
+        time:     2015-07-26 19:42:05
+        versions: ['<=3.0.6']
+reference: composer://codeigniter/framework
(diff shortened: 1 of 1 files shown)


codeigniter/framework/2016-07-26-1.yaml
```diff
@@ -0,0 +1,8 @@
+title: Critical SQL injection bug in the ODBC database driver
+link: http://forum.codeigniter.com/thread-65803.html
+cve: ~
+branches:
+    3.0-stable:
```
fabpot: should be just "3.0"

codeigniter/framework/2016-07-26-1.yaml
```diff
@@ -0,0 +1,8 @@
...
+        time:     2015-07-26 19:42:05
+        versions: ['<=3.0.6']
```
fabpot: Are version 1 and 2 also vulnerable?

fabpot:
Any news on this one?

codeigniter/framework/2016-07-26-1.yaml
kenjis: Probably yes, but nobody can install version 1 and 2 via Composer.

codeigniter/framework/2016-07-26-1.yaml
kenjis: I will fix it.

PR #169, opened 2016-09-08
Fix ranges in CakePHP range.

markstory:
I was off by one in one situation and all of 2.5.x is vulnerable.

Refs #116


markstory:
Thanks :+1:

PR #174, opened 2016-10-31
Add Gregwar/RST < v1.0.3

paragonie-scott:
There's a local file inclusion vulnerability in Gregwar/RST < 1.0.3. Even if you're using a newer version, you still have to go through [some steps](https://twitter.com/CiPHPerCoder/status/793076643468304384) to disable it properly.


fabpot:
Thank you @paragonie-scott.

PR #178, opened 2016-11-05
RCE in pygmentize

h4ckninja:


3f/pygmentize/2016-11-05.yaml
```diff
@@ -0,0 +1,9 @@
+title: Remote Code Execution
+link: https://github.com/dedalozzo/pygmentize/issues/1
+cve: ~
+branches:
+    master:
+        time: 2016-12-31 00:00:00
```
fabpot: This must be the date of the commit when it's merged, so I cannot merge this PR before it's fixed by the upstream package.

cebe: is there a way to make users of a library aware of a serious issue even if there is not fix yet?

3f/pygmentize/2016-11-05.yaml
```diff
@@ -0,0 +1,9 @@
...
+        versions: ['<=1.1']
```
stof: if it is not fixed upstream yet, such constraint is wrong, as you are not sure which version will include the patch.

micheal: Okay. I wasn't quite sure how to approach an application that has no fix yet, but @cebe was kind enough to help me get the upstream PR taken care of. I'll update this PR to reflect.

fabpot:
What's the status of this one? Still relevant?

h4ckninja:
The author responded to the PR, but it seems things stopped there. I just checked, and the fix was never merged in, leaving the library open to RCE.

cebe:
I have proposed a better fix in https://github.com/dedalozzo/pygmentize/pull/3

h4ckninja:
Ta muchly. In the same vein of discussion, I have another RCE I'm sitting on. I've notified the author of the quite popular library, but the repo itself has seen no updates in the last two years, and the author has not responded.

dedalozzo:
Merged, thank you.

3f/pygmentize/2016-11-05.yaml
stof: @h4ckninja can you update this to `<1.2` then, as it is the version of the release containing the fix ?

3f/pygmentize/2016-11-05.yaml
stof: @h4ckninja can you update the date now that the fix is merged ?

3f/pygmentize/2016-11-05.yaml
h4ckninja: I was a bit over-zealous in my fork pruning a few months ago and removed this fork. I'll close this PR and create a new one tonight.

h4ckninja:
#213 was created as I derped on purging my forks a bit aggressively.

PR #183, opened 2016-12-05
simplesamlphp SSPSA 201612-01: Incorrect signature validation

jdreesen:
This has already beed added for `simplesamlphp/saml2` in #182 but SimpleSAMLphp itself is also affected.

### DIFF
--- a/simplesamlphp/simplesamlphp/201612-01.yaml
+++ b/simplesamlphp/simplesamlphp/201612-01.yaml
@@ -1,0 +1,7 @@
+title:     Incorrect signature verification
+link:      https://simplesamlphp.org/security/201612-01
+branches:
+    master:
+        time:     2016-12-02 16:23:00
+        versions: ['<1.14.10']
+reference: composer://simplesamlphp/simplesamlphp


marcomenzel:
I'm not sure about these kind of issues. I had a look into SimpleSAMLphp, in my opinion it is only effected because of the dependency to the saml2-package?!

fabpot:
If the security issue only comes from a dependency managed by Composer, the advisory should be removed. @marcomenzel Can you confirm?

marcomenzel:
I do. I'm really sure about that. 
Commit 27b7efb81ad8d0fecbba725c0004aacedcaae1c8 in simplesamlphp solves the issue/pr in  https://github.com/simplesamlphp/saml2/pull/81 

It seems like the simplesamlphp team is used to do it this way (patch version bump on dependency security issues). Propably to get some attention from simplesamlphp-endusers.

fabpot:
reverted, thanks @marcomenzel

jdreesen:
Oh, well. Sorry for the mistake :(

Ocramius:
@jdreesen mistakes happen - no need to be sorry :-)

PR #186, opened 2016-12-19
simplesamlphp SSPSA 201612-02: Incorrect signature validation

jdreesen:
This time I'm rather sure that the issue is in simplesamlphp directly and not in a dependency.
The commit with the fix should be: https://github.com/simplesamlphp/simplesamlphp/commit/a2326d75dd14accaac162dd2cb30aaefcc1f9205

Maybe @marcomenzel could verify this before merging?


marcomenzel:
looks good to me :+1: 
thanks for verifying my hidden thoughts, I didn't find the fixing commit last time I added SSPSA 201602-03.

PR #187, opened 2016-12-21
ZF2016-04: Potential remote code execution ...

marcomenzel:
…in zend-mail via Sendmail adapter.

Would be nice to know if ZF1 is also affected :-/.

### DIFF
--- a/zendframework/zend-mail/ZF2016-04.yaml
+++ b/zendframework/zend-mail/ZF2016-04.yaml
@@ -1,0 +1,28 @@
+title:     Potential remote code execution in zend-mail via Sendmail adapter
+link:      https://framework.zend.com/security/advisory/ZF2016-04
+branches:
+    2.0.x:
+        time:     2016-12-19 22:44:00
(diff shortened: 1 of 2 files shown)


stof:
Looking at the code, ZF1 seems to be affected as well (and not patched as it is EOLed). @weierophinney can you confirm ?

weierophinney:
@stof and @marcomenzel — we reported the issue exactly as it was reported to us, which was specifically against the v2 zend-mail component. If you believe there may be a vulnerability against ZF1, please [use our security reporting mechanism to report the issue](https://framework.zend.com/security). As @stof notes, ZF1 is EOL, which means that we will not be issuing any public patches if we do determine it is vulnerable; however, we do provide patches for Zend Server customers when they report security issues to us.

jaydiablo:
This is discussed more here: https://github.com/zendframework/zf1/pull/733

As mentioned in the comment there, double quotes are stripped from those headers earlier in the flow.  I couldn't get the tests from ZF2 (adapted to ZF1) to exhibit the same behavior.

PR #188, opened 2016-12-26
CVE-2016-10033: RCE in PHPMailer

marcomenzel:


phpmailer/phpmailer/CVE-2016-10033.yaml
```diff
@@ -0,0 +1,8 @@
+title:     Remote Code Execution
+link:      https://github.com/PHPMailer/PHPMailer/releases/tag/v5.2.18
+cve:       CVE-2016-10033
+branches:
+    5.x:
+        time:     2016-12-23 23:40:00
+        versions: ['>=5.0.0', '<5.2.18']
```
cebe: this is not true, all versions < 5.0 are also affected. At least 2.0.4 is, which I found on one of my servers. we should not set a lower limit here.

marcomenzel: It should bei fine, because this repository only handles packages and their versions which are distributed by composer.

cebe: Okay, was not aware that only versions >=5 are available via composer. Fine then.

Federkun:
Wait for it, https://legalhackers.com/advisories/PHPMailer-Exploit-Remote-Code-Exec-CVE-2016-10045-Vuln-Patch-Bypass.txt

marcomenzel:
Hm, since this issue got an own cve identifier, I think this will get an own advisory in here

PR #189, opened 2016-12-28
CVE-2016-10045: RCE in PHPMailer

Federkun:


marcomenzel:
looks good to me :+1:

DavidPrevot:
The lower bound seems wrong, since as I understand it, the issue was introduced by the initial CVE-2016-10033 fix, aka 5.2.19.

Federkun:
More than a  "we introduced a new bug on 5.2.19" is a "the patch that we released on 5.2.18 was not enough". I'd keep it as it is.

DavidPrevot:
Let me try to rephrase what I understood: the issue tracked by CVE-2016-10045 was introduced by the incomplete CVE-2016-10033 fix in 5.2.19.

Federkun:
Yeah, that's why I would keep `>=5.0.0`, as every version until 5.2.20 are affected by this CVE.

cebe:
As already mentioned in #188 thinking about it more, adding a lower bound does not make sense here as all versions even before 5.0.0 are vulnerable. Even if these security advisories are targeted at composer packages, people reading this should not feel safe if they are using versions < 5.0.0.

PR #190, opened 2016-12-29
Add CVE-2016-10074: RCE in SwiftMailer

jmfontaine:


xabbuh:
By the way, shouldn't we list versions before 5.0 as affected too?

fabpot:
Ok, so quite a few people raced to get this one in :) I prefer my version as it is more specific. The issue is only with the mail transport... and hopefully, nobody is using it. see #192

jmfontaine:
@fabpot Actually, I came across the vulnerability disclosure last night and found out it was missing in this database earlier this morning. I found out a fix was released while checking the affected versions.

I don't mind whose PR is accepted as long as it's added to the database. :)

PR #193, opened 2017-01-06
Add Flow-SA-2017-01

kdambekalns:


stof:
this looks wrong to me. The RCE is not in neos/swiftmailer but in swiftmailer/swiftmailer, which is its dependency. And the RCE is swiftmailer/swiftmailer is already referenced, meaning that people are already receiving the info (your PR would make them get the same issue twice).

The flow team had to make a new release not to fix the RCE, but to allow users to install the patched version of swiftmailer (because they use exact version requirements)

kdambekalns:
Which is what our new release does (raise the required version of swiftmailer/swiftmailer). So… what should I do exactly? Change the title?

marcomenzel:
@kdambekalns release of neos/swiftmailer is the soultion for the before affected versions, but the library itself isn't affected. So there should be no issue mentioned in this repositoy.
It's some kind of special, because it is not sufficient to just do an `composer update` on neos/swiftmailer (<5.4.5), while there are strict version dependencies to swiftmailer/swiftmailer.

kdambekalns:
Sorry, I still don't get that. We require a vulnerable version. We fixed that dependency and released a new version. So this is a security release of our package. I think this should be contained here.

a) Do you agree?
b) If yes, what should I fix/change?

Also, you mean with

> it is not sufficient to just do an composer update on neos/swiftmailer (<5.4.5)

that with a fixed dependency to neos/swiftmailer the underlying issue will not be fixed? Right, even more important to let people know about the new version and the issue it fixes (even if transitively).

stof:
@kdambekalns the checker checks all composer packages used in the project. So it already detects the swiftmailer vulnerability when using the older version of neos/swiftmailer. Adding this advisory would report it twice instead.

As you are using exact versions, yes you had to make a release. But the reason you had to do it is not because your package itself is vulnerable. It is because your package was forbidding users to update swiftmailer to fix the vulnerability warning they were already getting since 8 days (when the advisory was added in #192 for swiftmailer).
This is why it should not be listed in this package.

Advocating the release as a security release on your website is fine of course. But people using this advisory database and the tooling based on it already know about it since 1 week (they were just unable to fix it before your release). That is, unless you built your own tooling based on this database but without taking all packages of the project into account of course, but this is not something this database want to solve (we cannot list all dependant packages as affected by each vulnerability here, especially as it would become a nightmare for any package using range constraint, and would be totally unmaintainable)

kdambekalns:
Ok. Admittedly I have never used the checked and was largely unaware of it. For me this was just a database I can follow in my RSS reader. ;)

PR #195, opened 2017-01-11
Added TYPO3-CORE-SA-2017-001 for TYPO3

derhansen:


marcomenzel:
Mh. Not that easy.
Version 6.2.0 to 6.2.29 are directly affected.
7.6.x and 8.x are not effected directly, you could easily fix the issue by updating swiftmailer through composer. I would remove them :)

xabbuh:
I agree with @marcomenzel. If those versions are not directly affected, there is no need to list them here. People being vulnerable due to an affected version of Swiftmailer will get a warning anyway because this repository includes the Swiftmailer entry too.

derhansen:
Agree, thanks for the feedback @marcomenzel. I updated the PR.

PR #198, opened 2017-01-24
shopware january 2017 advisory

marcomenzel:


fabpot:
Thanks you @marcomenzel

PR #206, opened 2017-03-02
Add PHP CodeSniffer advisory 2017-03-01

klausi:


squizlabs/php_codesniffer/2017-03-01.yaml
```diff
@@ -0,0 +1,12 @@
+title:     Arbitrary shell execution
+link:      https://github.com/squizlabs/PHP_CodeSniffer/releases/tag/2.8.1
+cve:       ~
+branches:
+    # The 1.x branch is unsupported, mark all versions as vulnerable.
+    1.x:
+        time:     2017-01-03 22:17:45
+        versions: ['>=1.0.0', '<1.99.99']
```
Ocramius: Just use 2.0 as upper boundary

PR #212, opened 2017-05-09
Add Artax cookie leakage

kelunik:
* https://twitter.com/asyncphp/status/862033196824842246
* https://github.com/amphp/artax/commit/accdadaf78f7a43305c3a97d6a964bbc550a555d

### DIFF
--- a/amphp/artax/2017-05-09.yaml
+++ b/amphp/artax/2017-05-09.yaml
@@ -1,0 +1,11 @@
+title:     Cookie leakage to wrong origins and non-restricted cookie acceptance
+link:      https://github.com/amphp/artax/releases/tag/v2.0.6
+cve:       ~
+branches:
+    2.x:
+        time:     2017-05-09 19:42:02
+        versions: ['>=2', '<2.0.6']
+    1.x:
+        time:     2017-05-09 19:42:02
+        versions: ['<1.0.6']
+reference: composer://amphp/artax


amphp/artax/2017-05-09.yaml
```diff
@@ -0,0 +1,11 @@
+title:     Cookie leakage to wrong origins and non-restricted cookie acceptance
+link:      https://twitter.com/asyncphp/status/862033196824842246
+cve:       ~
+branches:
+    master:
+        time:     2017-05-09 19:42:02
+        versions: ['>=2', '<2.0.6']
+    1.x:
+        time:     2017-05-09 19:42:02
+        versions: ['>=1', '<1.0.6']
```
Ocramius: I assume pre-1.0 versions are also affected?

amphp/artax/2017-05-09.yaml
```diff
@@ -0,0 +1,11 @@
+title:     Cookie leakage to wrong origins and non-restricted cookie acceptance
+link:      https://twitter.com/asyncphp/status/862033196824842246
```
stof: I would rather link to https://github.com/amphp/artax/releases/tag/v2.0.6 which gives some details about the vulnerability

amphp/artax/2017-05-09.yaml
kelunik: I assume, too. Should I just drop `>=1`?

How should I add the `amp_v2` branch? Fixed that just now, but it doesn't have any released version yet.

amphp/artax/2017-05-09.yaml
kelunik: Fixed locally, will push once @Ocramius responds to the other comment.

amphp/artax/2017-05-09.yaml
Ocramius: > I assume, too. Should I just drop `>=1`?

I would suggest dropping the lower boundary, yes

 > How should I add the amp_v2 branch? Fixed that just now, but it doesn't have any released version yet.

Not relevant then - doesn't need to be mentioned

amphp/artax/2017-05-09.yaml
kelunik: Pushed.

amphp/artax/2017-05-09.yaml
kelunik: Changed. How are branch names handled? Do they actually matter? Because `master` will change to `2.x` soon.

Ocramius: They don't really matter that much, yeah.

amphp/artax/2017-05-09.yaml
stof: looks like you have not pushed the updated code

kelunik: Indeed, forgot to `git add`.

amphp/artax/2017-05-09.yaml
fabpot: Changing `master` to `2.x` is actually a good idea.

kelunik:
@fabpot @Ocramius Could you merge?

fabpot:
Can you change master to 2.x?

kelunik:
Done.

PR #213, opened 2017-05-16
RCE in pygmentize

h4ckninja:


3f/pygmentize/2017-05-15.yaml
```diff
@@ -0,0 +1,8 @@
+title: Remote Code Execution
+link: https://github.com/dedalozzo/pygmentize/issues/1
+cve: ~
+branches:
+    master:
```
fabpot: should be 1.x

h4ckninja: https://github.com/FriendsOfPHP/security-advisories/pull/213/commits/ab619eed6f6aebe63345c78d2de7cae2b928d435 should fix this.

PR #226, opened 2017-09-06
EZSA-2017-005: XSS issue in search

glye:
Hi! This is eZ Systems' first PR here. We want to use this repo actively from now on. Because of the delay in getting CVEs we intend to publish issues here as soon as we make them public on our end, and update them later when we get the CVEs.

http://share.ez.no/community-project/security-advisories/ezsa-2017-005-xss-issue-in-search

### DIFF
--- a/ezsystems/ezpublish-legacy/2017-08-22-1.yaml
+++ b/ezsystems/ezpublish-legacy/2017-08-22-1.yaml
@@ -1,0 +1,10 @@
+title:     EZSA-2017-005 XSS issue in search
+link:      http://share.ez.no/community-project/security-advisories/ezsa-2017-005-xss-issue-in-search
+branches:
+    5.4.x:
+        time:     2017-05-26 15:23:00
+        versions: ['>=5.4.0','<5.4.10']
+    5.3.x:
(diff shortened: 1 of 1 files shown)


fabpot:
@glye Your strategy sounds good! Thanks

ezsystems/ezpublish-legacy/2017-08-22-1.yaml
```diff
@@ -0,0 +1,10 @@
+title:     EZSA-2017-005 XSS issue in search
+link:      http://share.ez.no/community-project/security-advisories/ezsa-2017-005-xss-issue-in-search
+branches:
+    5.4.x:
+        time:     2017-05-26 15:23:00
+        versions: ['>=5.4.0','<5.4.10']
+    5.3.x:
+        time:     2017-08-18 12:27:00
+        versions: ['>=5.3.0','<5.3.12.1']
```
stof: is 5.2 fine ? If no, I would remove this lower bound

glye: @stof 5.2 and below are not distributed via composer, so not relevant for `security-advisories`.

PR #229, opened 2017-09-11
EZSA-2017-006: Information disclosure in backend content tree menu

glye:
Ref: http://share.ez.no/community-project/security-advisories/ezsa-2017-006-information-disclosure-in-backend-content-tree-menu


glye:
Feck, forgot the validator. Version format issue.

glye:
Fixed, reopening if Github will let me.

PR #231, opened 2017-09-11
20170911 typo3 sa september

pixelbrackets:
Fix link to TYPO3-Security Announcement 2017-004


fabpot:
Thanks for the follow-up.

PR #232, opened 2017-09-11
[FEATURE] TYPO3: Add missing SA and versions

pixelbrackets:
Add missing security announcement TYPO3-CORE-SA-2017-002.

Also fix the version constraints in the existing security
announcements of 2017 (add some missing mayor release versions).


fabpot:
Thank you!

PR #233, opened 2017-10-25
add simplesamlphp/simplesamlphp SSPSA 201710-01

ocramleznem:


simplesamlphp/simplesamlphp/201710-01.yaml
```diff
@@ -0,0 +1,13 @@
+title:     Signature validation bypass
+link:      https://simplesamlphp.org/security/201710-01
+branches:
+    1.12.x:
+        time:     2017-10-25 09:54:00
+        versions: ['>=1.12.0', '<1.13.0']
+    1.13.x:
+        time:     2017-10-25 09:54:00
+        versions: ['>=1.13.0', '<1.14.0']
```
cebe: it does not seem to be fixed in 1.13 and 1.12 branches, so adding a time here would not be correct as far as I see:

> https://github.com/FriendsOfPHP/security-advisories#contributing
> time: The date when the security issue was fixed (most of the time the date of the commit that fixed the issue (2012-08-27 19:17:44) -- this information must be as accurate as possible as it is used to determined if a software is affected or not;

marcomenzel: I'm not really sure about that. 
I thought this has to be done for a reason I didn't question. I have seen this in all advisories in this package. 
@fabpot any hints on this. Actually I'm totally with @cebe about this.

marcomenzel: #242 related

fabpot:
I've updated this PR to removed the time for versions that are not maintained anymore.

PR #239, opened 2017-11-17
add new Symfony security advisories

xabbuh:


xabbuh:
@fabpot Can we specify a range of versions before a stable release (`[3.4.0,<3.4.0-BETA5]` did not look right to me)?

symfony/form/CVE-2017-16790.yaml
```diff
@@ -0,0 +1,23 @@
+title:     "CVE-2017-16790: Ensure that submitted data are uploaded files"
+link:      http://symfony.com/blog/cve-2017-16790-ensure-that-submitted-data-are-uploaded-files
+cve:       CVE-2017-16790
+branches:
+    2.7.x:
+        time:     2017-11-16 15:14:44
+        versions: ['>=2.7.0', '<2.7.37']
```
nicwortel: Shouldn't this include 2.7.37? According to http://symfony.com/blog/cve-2017-16790-ensure-that-submitted-data-are-uploaded-files, the affected versions include 2.7.37, and it is fixed in 2.7.38. (this goes for all other versions in this PR as well, I guess)

xabbuh: You are right. Thank you for noticing. Should be fixed in all files.

fabpot:
@xabbuh We should not list beta versions. By definition of being beta, they should not be used in production and so, we should not care about sec issues for them.

PR #241, opened 2017-11-20
Add advisories for (abandoned?) projects with unfixed vulnerabilities.

paragonie-scott:
* https://github.com/Cosmicist/AsymmetriCrypt/issues/4
* https://github.com/pagarme/pagarme-php/issues/29
* https://github.com/stormpath/stormpath-sdk-php/issues/132
* https://github.com/titon/framework/issues/93

### DIFF
--- a/asymmetricrypt/asymmetricrypt/2017-11-20.yaml
+++ b/asymmetricrypt/asymmetricrypt/2017-11-20.yaml
@@ -1,0 +1,8 @@
+title:      Padding Oracle Vulnerability in RSA Encryption
+link:       https://github.com/Cosmicist/AsymmetriCrypt/issues/4
+cve:        ~
+branches:
+    0.x:
+        time:     ~
+        versions: ['>=0.0.0', '<9.9.99']
+reference: composer://asymmetricrypt/asymmetricrypt
--- a/pagarme/pagarme-php/2017-11-20.yaml
+++ b/pagarme/pagarme-php/2017-11-20.yaml
@@ -1,0 +1,8 @@
+title:      Padding Oracle Vulnerability in RSA Encryption
+link:       https://github.com/pagarme/pagarme-php/issues/29
+cve:        ~
+branches:
+    3.x:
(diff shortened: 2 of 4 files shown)


asymmetricrypt/asymmetricrypt/2017-11-20.yaml
```diff
@@ -0,0 +1,8 @@
+title:      Padding Oracle Vulnerability in RSA Encryption
+link:       https://github.com/Cosmicist/AsymmetriCrypt/issues/4
+cve:        ~
+branches:
+    0.x:
+        time:     2017-01-19 13:03:00
```
fabpot: Time is the time when the package was fixed. Not sure what to do here as there is no fix AFAIU. We need to have a new convention. I would suggest to keep the time empty.

marcomenzel: Propably `~` should be used. 
I just saw it a few times for missing cve-identifiers (like here).

paragonie-scott: Okay, I'll send a superseding commit right now, then.

paragonie-scott: Looks like it's failing because `time` is required. Hmm.

fabpot: Indeed, it's not supported yet by the validator and the server. So, I will need to make some adjustments before merging this PR.

paragonie-scott: What if we relax that constraint but only if another key is set (e.g. `unfixed: TRUE`)

paragonie-scott:
If #242 is accepted, this should work now. :)

paragonie-scott:
This should be good to go, then.

fabpot:
Indeed, thanks for the pull request.

PR #242, opened 2017-11-20
Relax time constraint

paragonie-scott:
See #241 discussion

If `unfixed` is specified for a branch, this allows `time` to be optional and an upper bound to be omitted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,8 +71,9 @@ Contributing security advisories is as easy as it can get:
 
           * `time`: The date when the security issue was fixed (most of the
             time the date of the commit that fixed the issue (`2012-08-27
-            19:17:44`) -- this information must be as accurate as possible as
+            19:17:44`)) -- this information must be as accurate as possible as
             it is used to determined if a software is affected or not;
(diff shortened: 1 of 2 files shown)


paragonie-scott:
If `unfixed` is specified for a branch, this allows `time` to be optional and an upper bound to be omitted.

cebe:
skipping the upper bound on a branch makes it harder to generate version constraints for https://github.com/Roave/SecurityAdvisories

paragonie-scott:
Oh, that's true. I'll undo that piece.

stof:
the upper bound is always necessary, otherwise it will match all versions (it is not only for Roave, but also for the Sensiolabs checker). You could say `1.x` as branch with a constraint from `2.2.5 to 2.3.7` if you want (even though it would not make sense)

stof:
Or, we could allow to have unfixed and no upper bound if you don't have any other range with a higher set of bounds (meaning that all upper versions are affected, useful for unmaintained packages)

fabpot:
Closing in favor of #255

PR #253, opened 2018-02-18
SQL injection vulns found in all versions of Propel in LIMIT statement

halfer:
I have prepared advisory files to draw attention to this vulnerability. I've discussed raising this advisory on the [issue ticket for Propel 2](https://github.com/propelorm/Propel2/issues/1463).

I'm happy to make any changes suggested by maintainers here.

Once this is merged I will also create some test projects for Propel1 (unmaintained) and Propel2 (current). Propel3 does not have any releases yet.


propel/propel/2017-02-14.yaml
```diff
@@ -0,0 +1,8 @@
+title:     SQL injection possible with limit() on MySQL
+link:      https://github.com/propelorm/Propel2/issues/1463
+cve:       ~
+branches:
+    "2.0":
+        time:     2018-02-16 13:38:00
+        versions: ['>=2', '<=3']
```
Ocramius: 2.x doesn't have any releases at all: this range is empty (alpha versions are not covered)

halfer: Ah, a contributor on my linked above suggested that `>=2` would include alpha versions. Is that not correct? If not, is there an expression I can use to cover everything in the 2.x series - would `2.*` work?

halfer:
Hmm, regarding the Travis build failure, is it best if I remove the Propel3 report? No releases have been made here, not even alpha. The project is somewhat stalled, even though they probably have a fair-sized legacy user-base.

propel/propel/2017-02-14.yaml
Ocramius: Yeah, `2.0.0` is after `2.0.0-alpha1`, so it is not in the range. You can write in `>=2.0.0-alpha1`.

propel/propel3/2017-02-14.yaml
```diff
@@ -0,0 +1,8 @@
+title:     SQL injection possible with limit() on MySQL
+link:      https://github.com/propelorm/Propel3/issues/73
+cve:       ~
+branches:
+    "3.0":
+        time:     2018-02-16 13:38:00
+        versions: ['>=3', '<4']
+reference: composer://propel/propel3
```
fabpot: I think this file should be removed as the package does not exist on Packagist, so this file will never match any `composer.lock` file.

stof:
File names should be `2018-02-14`, not `2017-*`

propel/propel/2017-02-14.yaml
```diff
@@ -0,0 +1,8 @@
+title:     SQL injection possible with limit() on MySQL
+link:      https://github.com/propelorm/Propel2/issues/1463
+cve:       ~
+branches:
+    "2.0":
+        time:     2018-02-16 13:38:00
+        versions: ['>=2.0.0-alpha1', '<3']
```
stof: this has actually been merged, so the upper bound should be `<= 2.0.0-alpha7`. If they do a new release of Propel 2, it will be unaffected.

Ocramius: *if* - in that case, we'll update these definitions. For now, everything in the released versions is affected.

stof: thing is, the commit is already merged in Prople 2. So we know that 2.0.0-alpha7 will always be the last affected version (unless they do crazy stuff by tagging a release on an older commit before the security fix, but then they would be insane)

marcj: > unless they do crazy stuff by tagging a release on an older commit before the security fix

which we won't.

propel/propel3/2017-02-14.yaml
marcj: Yes, v3 isn't released nor made public, so there's no point in tagging this version as vulnerable.

propel/propel/2017-02-14.yaml
Ocramius: @marcj don't get me wrong, I have full trust in the maintainers' good faith, but the current knowledge is that no releases in the `2.x` range are unaffected, so this repository should simply reflect that. Range can be adapted after a release is there :+1:

marcj: https://github.com/propelorm/Propel2/releases/tag/2.0.0-alpha8 has been tagged 1minute before you posted it. :)

Ocramius: :+1: awesome, then this can be changed

halfer:
Thanks for all the comments, everyone. Summary of changes:

* Propel2 now has a specific end version
* Propel3 is removed, as it does not have a Composer release
* Filenames are renamed from `2017-*` to `2018-*`

PR #255, opened 2018-02-21
Allow time to be null when the issue is not fixed yet

fabpot:
Alternative to #242. I don't think we need the `unfixed` entry. Having a `time` set as `null` is semantically enough.

/cc @paragonie-scott @stof

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,7 +73,7 @@ Contributing security advisories is as easy as it can get:
             issue is not fixed yet (most of the time, the date of the commit
             that fixed the issue in the following format `2012-08-27 19:17:44`)
             -- this information must be as accurate as possible as it is used
-            to determined if a software is affected or not;
+            to determine if a project is affected or not;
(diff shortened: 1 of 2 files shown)


stof:
Not making any change to the validator looks weird to me. Does it work ?

And for unfixed vulnerabilities, we also need to be able to mark all versions as impacted (so no upper bound)

fabpot:
The validator does not care about the value of the `time` entry, it just validates that it's present.

For the upper bound, it's just a matter of using a very high version, I don't think we need anything special here. As done in #241 (even if I would use a version a bit higher than `9`, let's say `999`).

fabpot:
I think trying to keep the current semantic as much as possible is important here.

stof:
but setting to `null` will not be considered as there when checking with `isset`

fabpot:
@stof indeed, fixed

PR #257, opened 2018-02-21
Add advisory for Parsedown

aidantwoods:
Parsedown has had multiple XSS issues for a long time when using its `->setMarkupEscaped(true)` option (e.g. being able to break out of the AST representation by adding quotes in link ~~addresses~~ titles). All of these issues are fixed in https://github.com/erusev/parsedown/pull/495, but this has remained open for almost a year now, and there appears to be no interest in merging it any time soon from the repository owner (I digress).

There might be some debate as to whether XSS issues in a markdown parser can actually be considered security issues (since markdown itself permits HTML). To address this I would present the following:
* The Parsedown Wiki has for a long time [stated](https://github.com/erusev/parsedown/wiki/Tutorial:-Get-Started/f9d881da38d8ef737ffc0c0a140b124ba3367904) that HTML can be escaped using the `->setMarkupEscaped(true)` configuration:
  > ```php
  >   echo Parsedown::instance()
  >      ->setMarkupEscaped(true) # escapes markup (HTML)
  >      ->text("<div><strong>*Some text*</strong></div>");
  >
  >   # Output:
  >   # <p>&lt;div&gt;&lt;strong&gt;<em>Some text</em>&lt;/strong&gt;&lt;/div&gt;</p>
  >```

  (Parsedown does not do this properly)

* Attempts at escaping HTML appear throughout the code, but are not sufficient, or are applied in some places but missed in analogous ones (e.g. https://github.com/erusev/parsedown/pull/495#issuecomment-340721927)

### DIFF
--- a/erusev/parsedown/2017-05-01.yaml
+++ b/erusev/parsedown/2017-05-01.yaml
@@ -1,0 +1,7 @@
+title:     Cross-Site Scripting
+link:      https://github.com/erusev/parsedown/pull/495
+branches:
+    1.x:
+        time:     ~
+        versions: ['<=1.6.4']
+reference: composer://erusev/parsedown


fabpot:
I've read the thread there and your description here. I don't want to be the only one who takes the decision. From my point of view, I think this should be merged. But let's see what others think. /cc @stof @nicolas-grekas @ircmaxell @paragonie-scott (I could ping more here, but that's a good start:))

Ocramius:
Seems pretty much clear that until a release with that patch in, that
package is unusable.

That can be changed once the release has happened, but nobody should
install or run parsedown until then.

paragonie-scott:
I'd be wary of any project that sits idle while security fixes lurk unmerged for months/years, just in case a critical 0day is found being exploited in the wild. The fact that the vulnerability in question is unfixed right now, I would say merge it.

If the fix is not included in the next release, I would keep bumping up the version constraint until the fix is implemented.

Strong disagree with "parser doesn't imply sanitizer" if your parsed output can introduce XSS vectors where none existed previously. (e.g. `htmlentities()` then Parsedown -> XSS anyway).

aidantwoods:
> If the fix is not included in the next release, I would keep bumping up the version constraint until the fix is implemented.

To be honest, and it kinda sucks to say, but I wouldn't be surprised if the next version didn't include the fix. The only things that really seem to get merged any more are one-liners, or travis config changes. This just one of many issues that have fixes in PRs, literally just waiting to be merged. I wonder whether the project is just dead from lack of maintenance tbh.

taylorotwell:
Eh, since the library does appear to make some explicit claim about escaping with its `setMarkupEscaped` method maybe it does warrant inclusion here.

Ocramius:
As per comment on https://github.com/Roave/SecurityAdvisories/issues/44, it is possible to generate arbitrary XSS anyway

garrettrayj:
The Parsedown readme states that a sanitizer should also be used when dealing with user generated content. The Symfony example includes HTML purifier. Escaped markup isn't necessarily sanitized and I don't see claims of sanitization anywhere.

Given the developer provides warnings and examples of how to work with risky content I think this advisory was a bad call.

Ocramius:
May I remind you of https://github.com/rails/rails/issues/5228 please?

Out of the box, that package is insecure.
The example at https://github.com/FriendsOfPHP/security-advisories/pull/257#issuecomment-368591110 shows that an external escaper is not used by experienced users.

aidantwoods:
> Escaped markup isn't necessarily sanitized and I don't see claims of sanitization anywhere.

Something that appears to repeatedly keep being neglected is that the parser fails at both in its current release version. **The parser does make claims about escaping, and it fails at doing it**. The fact that it is possible to generate arbitrary HTML when a mode called `setMarkupEscaped` is enabled is very clearly a bug with security consequences. 

Additionally the note in the readme you reference was only added after https://github.com/erusev/parsedown/pull/495 was opened because the maintainer has not had time to review the PR (this is in the thread over there).

Also, something else to note. This isn't some hypothetical problem where a fix is complicated to arrive at, or being a huge burden to implement. There is a fix literally sitting in a PR. The work is already done. This entire situation can be resolved by clicking a merge button. It's that easy.

paragonie-scott:
@garrettrayj 

>  I think this advisory was a bad call

I disagree wholeheartedly that this advisory was a bad call. It does introduce an XSS vulnerability where one did not previously exist, which is not also found in other Markdown parsers. If I worked for MITRE, I would assign it a CVE based on that alone.

I think what many in the Laravel community (include @taylorotwell's initial tweet) meant to say is rather, "This is inconvenient." Which it is! But I want you all to keep something very important in mind.

### Where there's smoke, there's fire.

If the maintainer would let this vulnerability sit around for months, with a patch available, for a novel vulnerability that only affects their library, then they clearly aren't up to the task of maintaining this library for the size of the community that depends on it. (That's no reflection on their character, life happens to us all!)

Anyone who thinks this advisory was a bad call because it makes life inconvenient for them should do one or more of these things:

1. Help with the maintenance of the Parsedown library.
2. Help with the maintenance of a Laravel fork of the Parsedown library.
3. Pay money, instead of time, to another developer so they can do 1 or 2 in your stead.

I don't think protesting this security advisory is a good use of anyone's time. It is *emphatically* a vulnerability. If anyone doesn't think this should "count" as a vulnerability, I'd love to [take a look](https://paragonie.com/blog/2017/06/why-you-want-paragon-initiative-enterprises-audit-your-code) at code you believe is secure.

garrettrayj:
@Ocramius @aidantwoods @paragonie-scott 

All excellent points. The fact that the note in the readme was added after the PR was made is especially damning evidence that I missed. My mind is changed, security advisory is warranted.

taylorotwell:
I agree that Parsedown's claim at escaping and it's failure to properly do so warrant its inclusion here, unfortunately. If it had made no claim to offer escaping at all I don't think it would warrant inclusion.

dregad:
For the record, @aidantwoods's PR erusev/parsedown#495 was (finally) merged this morning and [version 1.7.0 has been released](https://github.com/erusev/parsedown/releases/tag/1.7.0).

Ocramius:
@dregad can you send a patch to change this to `<1.7.0` and with an updated link (if any)?

PR #261, opened 2018-03-01
Adjust report version range

leonampd:
Looking at the code I see that the security issue about
[pagarme/pagarme-php](https://github.com/pagarme/pagarme-php) reported
in #241 only appears on versions below 3.x. Since the stable version (on
branch V3) doesn't have the vulnerability this PR aims to fix the
version range.

### DIFF
--- a/pagarme/pagarme-php/2017-11-20.yaml
+++ b/pagarme/pagarme-php/2017-11-20.yaml
@@ -2,7 +2,7 @@ title:      Padding Oracle Vulnerability in RSA Encryption
 link:       https://github.com/pagarme/pagarme-php/issues/29
 cve:        ~
 branches:
-    3.x:
+    2.x:
         time:     ~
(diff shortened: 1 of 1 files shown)


cebe:
> Since the stable version (on
> branch V3) doesn't have the vulnerability this PR aims to fix the
> version range.

can you give a reference to the code or commits, so we can check how it is fixed?

leonampd:
Hi @cebe hows it going?

The 3.x version is totally different from 2.x and 1.x. It's a new version written from scratch and you can check this on its first commit [here](https://github.com/pagarme/pagarme-php/tree/6e710eb0c57c80e07218f6f73995eaaadca78880).

Also the problem reported is about a RSA keys manipulation wich doesn't appear on 3.x at `Card` context as you can see in this [link](https://github.com/pagarme/pagarme-php/tree/V3/lib/Card). The card hash generation mentioned in #241 is done by the API which this library communicates.

PR #265, opened 2018-03-05
Create 201803-01.yaml

forevermatt:
I'm not sure I have the time exactly as it should be. The example I looked at didn't match to the second (it used `00` for the second, so I did as well), but someone familiar with submitting these should probably check that date/time value.

This security advisory also affected simplesamlphp/saml2, but unfortunately I don't have time to submit a file for that path at the moment.

### DIFF
--- a/simplesamlphp/simplesamlphp/201803-01.yaml
+++ b/simplesamlphp/simplesamlphp/201803-01.yaml
@@ -2,6 +2,6 @@ title:     Incorrect signature verification
 link:      https://simplesamlphp.org/security/201803-01
 branches:
     1.15.x:
-        time:     2018-03-02 16:02:00
+        time:     2018-03-02 15:02:24
         versions: ['<1.15.4']
 reference: composer://simplesamlphp/simplesamlphp
--- a/simplesamlphp/simplesamlphp/201803-01.yaml
+++ b/simplesamlphp/simplesamlphp/201803-01.yaml
@@ -1,0 +1,7 @@
+title:     Incorrect signature verification
+link:      https://simplesamlphp.org/security/201803-01
+branches:
+    1.15.x:
+        time:     2018-03-02 16:02:00
+        versions: ['<1.15.4']
+reference: composer://simplesamlphp/simplesamlphp


simplesamlphp/simplesamlphp/201803-01.yaml
```diff
@@ -0,0 +1,7 @@
+title:     Incorrect signature verification
+link:      https://simplesamlphp.org/security/201803-01
+branches:
+    1.15.x:
+        time:     2018-03-02 16:02:00
```
fabpot: To answer your question, you should use the exact time (including seconds) in UTC.

stof:
and we also need the advisory for simplesamlphp/saml2 for the same issue

forevermatt:
@jaimeperez Would you check the `time` in this? It doesn't seem to match the examples of your previous submissions, but as far as I can tell it is the exact (with seconds) commit date, in UTC, of the commit where this vulnerability was fixed (aka. the commit tagged as `1.15.4`).

forevermatt:
@stof I assume you saw it already, but @jaimeperez has submitted (and gotten merged) the advisory for simplesamlphp/saml2 that you (correctly) mentioned would be needed:
https://github.com/FriendsOfPHP/security-advisories/blob/master/simplesamlphp/saml2/201803-01.yaml

jaimeperez:
@stof I provided an advisory for simplesamlphp/saml2 (#266), and it has just been merged.

@forevermatt the time should be 15:01, I believe. The relevant commit is the one where we bumped the dependency of the SAML2 library, and that was committed at 16:01 local time (again, if I'm not mistaken), which is CET. So that should be 15:01 UTC.

In any case, I'm a bit hesitant to include an advisory for SimpleSAMLphp itself, and that's actually why I didn't in my PR. The reason is that SimpleSAMLphp had nothing in its code related to the vulnerability. The actual issues were in XMLSecLibs and the SAML2 library, and the fix for it was done in the SAML2 library as well. We released a new version of SimpleSAMLphp because that's easier than telling all our users to do a `composer update`, but SimpleSAMLphp was vulnerable here by the transitive property. I don't know if there's a policy here for this kind of cases. Is it common to add advisories for software that **uses** a vulnerable dependency?

fabpot:
If the vulnerability is in a dependency, then we need an advisory on the dependency. Do you want me to revert this PR?

stof:
if it was only vulnerable transitively, it should indeed not have an advisory here. The composer.lock will contain the saml2 library, which will be checked properly (and then simplesamlphp/simplesamlphp 1.15.3 is not vulnerable if the latest simplesamlphp/saml2 version is used with it)

stof:
@fabpot reverting would make sense to avoid duplicate vulnerability reporting (both for simplesamlphp/simplesamlphp and simplesamlphp/saml2)

jaimeperez:
I think it's a good idea to revert, yes. As @stof says, the composer.lock will have the vulnerable version of the SAML2 library, so if someone tries to install SimpleSAMLphp by running `composer install` on a 1.15.2 release, it should report the issue, whereas `composer update` should work.

Thanks for the clarification guys!

fabpot:
reverted in #267

forevermatt:
Sorry for the mis-submission here. Thanks for cleaning that up.

PR #266, opened 2018-03-06
Add recent vulnerabilities on the simplesamlphp/saml2 library

jaimeperez:
Ditto.


fabpot:
@jaimeperez Thanks for the PR. Can you rename the files to `.yaml` instead of `.xml`?

jaimeperez:
Of course! What a silly mistake. Should be fixed now.

PR #275, opened 2018-03-20
Fix `time` in advisory for squizlabs/php_codesniffer < 2.8.1

jrfnl:
The date/time of the issue was in an incorrect format (`2017-01-03` vs `2017-03-01`).

As I was fixing this anyway, I checked the actual time of the commit fixing this which can be seen as the first public announcement of the vulnerability.

So the new time is based on when the fix was originally committed to PHPCS: https://github.com/squizlabs/PHP_CodeSniffer/commit/029305e754dde022c4c8ee48d392c16b61fffe2a

### DIFF
--- a/squizlabs/php_codesniffer/2017-03-01.yaml
+++ b/squizlabs/php_codesniffer/2017-03-01.yaml
@@ -4,9 +4,9 @@ cve:       ~
 branches:
     # The 1.x branch is unsupported, mark all versions as vulnerable.
     1.x:
-        time:     2017-01-03 22:17:45
+        time:     2017-02-26 22:15:53
         versions: ['>=1.0.0', '<2.0.0']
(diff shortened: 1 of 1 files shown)


fabpot:
What matters is the time of the merge commit.

jrfnl:
@fabpot Thanks for confirming that. That's what I've used now for this one. I will adjust my other PR.

fabpot:
@jrfnl Looks like you did not push.

jrfnl:
@fabpot For this PR, no change is needed. The time in the original PR I submitted is based on when the fix was merged into PHPCS, so is correct already.

I've pushed an amended commit to #274 to fix the time in that one.

fabpot:
Ok, understand. Thanks for confirming.

PR #279, opened 2018-03-31
Add Laravel weak comparison encryption failure exploit

byCedric:
As mentioned by Taylor Otwell, https://medium.com/@taylorotwell/laravel-security-release-5-6-15-and-5-5-40-56f1257933a0

### DIFF
--- a/illuminate/encryption/2018-03-30-1.yaml
+++ b/illuminate/encryption/2018-03-30-1.yaml
@@ -1,0 +1,35 @@
+title:     Exploit of encryption failure vulnerability
+link:      https://medium.com/@taylorotwell/laravel-security-release-5-6-15-and-5-5-40-56f1257933a0
+cve:       ~
+branches:
+    4.0.x:
+        time:     ~
+        versions: ['>=4.0.0', '<=4.0.11']
+    4.1.x:
+        time:     ~
+        versions: ['>=4.1.0', '<=4.1.31']
+    4.2.x:
+        time:     ~
+        versions: ['>=4.2.0', '<=4.2.22']
+    5.0.x:
+        time:     ~
+        versions: ['>=5.0.0', '<=5.0.35']
+    5.1.x:
+        time:     ~
+        versions: ['>=5.1.0', '<=5.1.46']
+    5.2.x:
+        time:     ~
+        versions: ['>=5.2.0', '<=5.2.45']
+    5.3.x:
+        time:     ~
+        versions: ['>=5.3.0', '<=5.3.31']
+    5.4.x:
+        time:     ~
+        versions: ['>=5.4.0', '<=5.4.36']
+    5.5.x:
+        time:     2018-03-30 13:26:38
+        versions: ['>=5.5.0', '<5.5.40']
+    5.6.x:
+        time:     2018-03-30 13:26:38
+        versions: ['>=5.6.0', '<5.6.15']
+reference: composer://illuminate/encryption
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
All vesions less than 5.5 are vulnerable actually, including 5.4, 5.3, 5.2, 5.1, 5.0. You'll have to check how far back in the code this goes. Laravel 4 and 3 may be affected too.

byCedric:
Oh snap, you are right. I thought I saw a commit of a year old containing this. Let me check it tonight, I will update it to the proper versions! Besides this, is the PR ok? I mean it's my first PR here, not sure if I'm doing it right 😄

laravel/framework/2018-03-30-1.yaml
```diff
@@ -0,0 +1,11 @@
+title:     Exploit of encryption failure vulnerability
+link:      https://medium.com/@taylorotwell/laravel-security-release-5-6-15-and-5-5-40-56f1257933a0
+cve:       ~
+branches:
+    5.5.x:
+        time:     2018-03-30 14:15:00
```
xabbuh: It seems to me that the commit fixing the issue is laravel/framework@28e53f23a76206fb130e9a54eb95aa3f010e79c9 which was made at 13:26.

laravel/framework/2018-03-30-1.yaml
```diff
@@ -0,0 +1,11 @@
...
+        versions: ['>=5.5.0', '<5.5.40']
+    5.6.x:
+        time:     2018-03-30 14:15:00
```
xabbuh: It seems to me that the commit fixing the issue is laravel/framework@886d261df0854426b4662b7ed5db6a1c575a4279 which was made at 13:24.

laravel/framework/2018-03-30-1.yaml
byCedric: Yep! You are totally right! I mistakenly reused the "time of announcement" or whatever you want to call it. I just updated this to match the exact commit date. Thanks for the check!

laravel/framework/2018-03-30-1.yaml
byCedric: Thanks again, [see this comment](https://github.com/FriendsOfPHP/security-advisories/pull/279/files/0033f23237a63b3bcb13c51955b047a0e0a7f35f#r178441281).

byCedric:
I dug through the history of Laravel's encryption. It brought back some good memories, especially with Laravel 4. 😄 And, like you said, it appears that no version ever (before the patched ones) checked the length of the IV parameter when decrypted. 

Note that the Laravel `4.x.x`, all the way up to Laravel `5.0.29`, is using mcrypt. But I'm guessing that doesn't affect the behaviour of this issue.

Also I couldn't find any references to Laravel `3.x` anymore. Because of this I cannot verify if this issue also applies to Laravel `3.x`, although I think its very likely.

illuminate/encryption/2018-03-30-1.yaml
```diff
@@ -0,0 +1,35 @@
...
+        versions: ['>=5.1.0', '<=5.1.46']
+    5.2.x:
+        time:     ~
+        versions: ['>=5.2.0', '<=5.2.45']
+    5.3.x:
+        time:     ~
+        versions: ['>=5.3.0', '<=5.3.31']
+    5.4.x:
+        time:     ~
+        versions: ['>=5.4.0', '<=5.4.36']
```
xabbuh: AFAIK you can simplify these entries:

```yaml
'5.x':
    time: null
    versions: ['>=5.0.0', '<5.5.40']
```

xabbuh: Note: There is no need for the explicit `null` here, but I don't know how to write the tilde character on my mobile. ;)

fabpot: no, it should not be simplified.

byCedric: I'd figure it would work different since [Laravel doesn't follow semver](https://twitter.com/taylorotwell/status/698181502215938048), so we have to be a little more explicit like this right?

PR #290, opened 2018-04-20
Use HTTPS whenever possible

fabpot:


shopware/shopware/2017-01-25.yaml
```diff
@@ -1,5 +1,5 @@
 title:     Remote Code Execution Vulnerability
-link:      http://en.community.shopware.com/_detail_1989.html
+link:      https://en.community.shopware.com/_detail_1989.html
```
Majkl578: DId you check if HTTPS works? Because this domain has invalid certificate for example.

fabpot: Yes, I've checked them all by hand. Don't know how this one does not work now :( Can you submit a PR to revert to http?

bcremer: /ping  @soebbing @janbuecker @ndzoesch

bcremer: You can use https://community.shopware.com/_detail_1989.html (without the .en. subdomain)

fabpot: ah, I remember that I've tried https://community... for another advisory, so I think I've just thought that all subdomains would be ok.

bcremer: It should be OK to omit the `en.` prefix for all these links as the detail pages are deep links to the english versions anyway.

fabpot: done in f9ad9c5

PR #292, opened 2018-04-22
add drupal advisory SA-CORE-2018-003

ocramleznem:


fabpot:
Thank you @ocramleznem

PR #296, opened 2018-05-09
Fix fuelphp.

inisik:
- The name has been fixed.
- I disagree with the version constraint change.


stof:
The previous PR included 2 files, you renamed only one.

And why do you disagree with the version constraint ? the page you linked says that the 2016 vulnerability affects versions since 1.1 and is fixed in 1.8.0.4 (which is confirmed by the github commit for that part, as it belongs to the 1.8.0.4 release).

and for the upper bound of the 2018 vulnerability, 1.8.0.4 is also affected AFAICT, but 1.8.0.4 is a higher version that 1.8.0 in semver (and so in composer).

inisik:
I was in the process, I am on phone.

inisik:
The version constraint dir 2018 is fine but a added a lower constraint to the 2016 one.

stof:
Well, as I said, the 1.8.0.4 release is not marked as affected by the `<=1.8.0` constraint you used currently.

Btw, the https://github.com/fuel/core/commit/59112c96d0a6f2b0ead6a57edd8ac465678bdcb0 commit which is referenced as being the fix for the Crypt vulnerability is not part of the 1.8.1 release either

inisik:
Just checked the source code for 1.8.1 and those releases does contain the fix so 1.8.1 is not affected. You can view the file at https://github.com/fuel/core/blob/1.8.1/classes/crypt.php. And for the other security issue the 1.8.0.4 version does contain the fix. You can view the files here https://github.com/fuel/core/blob/1.8.0.4/classes/image/imagemagick.php. Versions 1.8.0.3 and below affected. Ill update the version constraints to these.

PR #297, opened 2018-05-21
Add new issue in CakePHP

markstory:
We had a cross site scripting flaw in some of our development mode error pages.


fabpot:
Thank you @markstory

PR #299, opened 2018-05-29
Add Symfony's May 2018 advisories

thewilkybarkid:


fabpot:
@thewilkybarkid Great work! I think we also need to add the unmaintained versions of Symfony as they are affected as well by all these issues.

thewilkybarkid:
Pre 2.7? The advisories themselves don't mention them, but can add.

thewilkybarkid:
Hopefully now complete!

sroze:
Great. Seems accurate with what I started. Out of curiosity... did you automate this or did you manually update the file? 🤔

thewilkybarkid:
Manually. Bit annoying to do due to the repo splitting, but better than not being there. :smiley:

fabpot:
I would remove the beta versions, they are not covered.

symfony/ldap/CVE-2018-11407.yaml
```diff
@@ -0,0 +1,26 @@
...
+    3.3.x:
+        time:     2018-05-25 12:44:29
+        versions: ['>=3.3.0', '<3.3.17']
+    3.4.x:
+        time:     2018-05-25 13:17:43
+        versions: ['>=3.4.0', '<3.4.11']
+    4.0.x:
+        time:     2018-05-25 14:04:47
+        versions: ['>=4.0.0', '<4.0.11']
+reference: composer://symfony/ldap
```
xabbuh: this looks wrong to me as the `LdapBindAuthenticationProvider` class is part of the `symfony/security-core` package

xabbuh: CVE-2018-11407 must also be added to `symfony/security` then.

stof:
@thewilkybarkid the easier way would have been to remove the lower bound on the `2.7.x` range, to make it include all older versions.

thewilkybarkid:
The branches themselves would not have been covered though?

PR #304, opened 2018-07-03
Add SUPEE-10752 for Magento1

barryvdh:
See https://magento.com/security/patches/supee-10752


barryvdh:
I do want to note that magento1 isn't officially supported with composer, so this is only useful as a reference, not for actually blocking something (or checking your composer.lock file for that matter)

PR #305, opened 2018-07-03
Magento 2.2.5 and 2.1.14 Security update

barryvdh:
See https://magento.com/security/patches/magento-2.2.5-and-2.1.14-security-update


stof:
The blog post you linked contains multiple security issues, with different affected ranges. So should they be separate advisories ?

barryvdh:
I don't know, does that matter? 
Most are 'Magento 2.1 prior to 2.1.14, Magento 2.2 prior to 2.2.5', while some are M2.1 or M2.2 specific. But they all are fixed on the new version.

barryvdh:
Replaced by #306

PR #308, opened 2018-07-09
Add an information about CSRF vulnerability in Sylius

pamil:


xabbuh:
Should there be entries for `sylius/admin-bundle` and `sylius/resource-bundle` too?

pamil:
Yeah, I'll add AdminBundle as well, ResourceBundle isn't affected directly.

pamil:
Done!

Ocramius:
> Yeah, I'll add AdminBundle as well, ResourceBundle isn't affected directly.

Is the code of those directly affected, or just through the dependency?

sylius/admin-bundle/2018-07-09.yaml
```diff
@@ -0,0 +1,14 @@
...
+branches:
+    1.0.x:
+        time: 2018-07-08 23:47:50
+        versions: ['>=1.0.0', '<1.0.17']
+    1.1.x:
+        time: 2018-07-09 00:09:14
+        versions: ['>=1.1.0', '<1.1.9']
+    1.2.x:
+        time: 2018-07-09 00:26:07
+        versions: ['>=1.2.0', '<1.2.2']
```
xabbuh: the newest release is 1.2.1

xabbuh: In fact, this applies to the other branches as well. It seems that subtree split releases are not always done when releases for the main repo are tagged.

pamil: All subtree split packages will be tagged in up to ~15 minutes.

pamil:
@Ocramius AdminBundle is directly affected, ResourceBundle contains a new feature in a generic action, which validates a CSRF token (that action can also be used through REST API).

sylius/admin-bundle/2018-07-09.yaml
pamil: Should be already tagged.

fabpot:
Thank you @pamil

PR #313, opened 2018-08-02
Add CVE-2018-14773

BackEndTea:


BackEndTea:
Thank you @fabpot

fabpot:
added the same for symfony/symfony in efcc6ea

stof:
@fabpot shouldn't we also have an advisory for http-kernel for the other CVE ?

fabpot:
sure

PR #315, opened 2018-08-09
Add advisory for Laravel cookie serialization vulnerability

byCedric:
As mentioned in the [official docs](https://laravel.com/docs/5.6/upgrade#upgrade-5.6.30) and [Laravel News](https://laravel-news.com/laravel-5-6-30). The merged pull request containing the fix for `5.6.x` branch is [over here](https://github.com/laravel/framework/pull/25121). And the commit for the `5.5.x` branch is either [this one](https://github.com/laravel/framework/commit/c5dc9c870a8a1af6c7e827dd81bd15f0a1561cc4), the [parent of that one](https://github.com/laravel/framework/commit/6df8767c657b5b14bb8ac23b4858f9f8def0cb3c) or the [parent-of-the-parent](https://github.com/laravel/framework/commit/3716d145b60150c1f86c25b4004390a23d8328ea).

### DIFF
--- a/illuminate/cookie/2018-08-08-1.yaml
+++ b/illuminate/cookie/2018-08-08-1.yaml
@@ -1,0 +1,35 @@
+title:     Cookie serialization vulnerability
+link:      https://laravel.com/docs/5.6/upgrade#upgrade-5.6.30
+cve:       ~
+branches:
+    4.0.x:
+        time:     ~
+        versions: ['>=4.0.0', '<=4.0.11']
+    4.1.x:
+        time:     ~
+        versions: ['>=4.1.0', '<=4.1.31']
+    4.2.x:
+        time:     ~
+        versions: ['>=4.2.0', '<=4.2.22']
+    5.0.x:
+        time:     ~
+        versions: ['>=5.0.0', '<=5.0.35']
+    5.1.x:
+        time:     ~
+        versions: ['>=5.1.0', '<=5.1.46']
+    5.2.x:
+        time:     ~
+        versions: ['>=5.2.0', '<=5.2.45']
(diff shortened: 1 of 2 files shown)


BackEndTea:
I think this should also include the [illuminate/cookie](https://github.com/illuminate/cookie) repo.

byCedric:
@BackEndTea You are right, added it to `illuminate/cookie` too! Thanks, almost missed that one.

ottowayne:
I think this affects Laravel 4 too: https://github.com/laravel/framework/blob/4.2/src/Illuminate/Encryption/Encrypter.php#L98

Back then you weren't even able to tell the Encrypter not to use serialization. The switch wasn't introduced until 2 years ago: https://github.com/laravel/framework/commit/9725a8e7d0555474114f5cad9249fe8fe556836c

byCedric:
@ottowayne Ah, yeah, I was thinking "its a middleware in which the vulnerability exists, and that wasn't introduced before 5.x".

I think you are right here too, I will add the whole `4.x.x` range to `illuminate/cookie` too. It's using `illuminate/encryption` in the insecure way. I don't think we have to add this library too right? Because the issue is only scoped to deserializing cookies, or am I missing something?

Ocramius:
Only what actually causes a security issue should be added here

byCedric:
Ok, thanks, I guess this is it then! 😀

ottowayne:
Yes, I thought about illuminate/encryption, too. In my opinion it was unfortunately designed, but it doesn't have a security issue by itself.

I guess that developers might not have known that stuff that got thrown in there will be serialized and are using it to encrypt user input.

PR #319, opened 2018-09-15
Add CVE for openid/php-openid

legoktm:
Note that I explicitly didn't include CVE-2016-2049, which is still unfixed (https://github.com/openid/php-openid/issues/128).

### DIFF
--- a/openid/php-openid/CVE-2013-4701.yaml
+++ b/openid/php-openid/CVE-2013-4701.yaml
@@ -1,0 +1,8 @@
+title:     XML External Entity (XXE) issue
+link:      https://github.com/openid/php-openid/commit/625c16bb28bb120d262b3f19f89c2c06cb9b0da9
+cve:       CVE-2013-4701
+branches:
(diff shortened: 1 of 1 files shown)


fabpot:
Thank you @legoktm

stof:
if CVE-2016-2049 is not fixed yet, I think it should still be included (marked as unfixed). The fact that the vulnerability is not patched yet does not make projects using the package secure.

legoktm:
@stof how would I indicate that it's unfixed? If I try putting `versions: ['*']` in the yaml file, the validator complains that there needs to be an upper bound. But we can't put an upper bound in, since we have no idea when the maintainers are going to fix the issue.

stof:
@fabpot you added support for putting `null` as `time` when no fix is available, but there does not seem to be a way to specify a valid config for `versions` for such case. What is the expected way ?

Ocramius:
Not specifying any range at all (`[]`) would be more correct

PR #328, opened 2018-10-21
Drupal Core - Multiple Vulnerabilities - SA-CORE-2018-006

ocramleznem:
from https://www.drupal.org/sa-core-2018-006


fabpot:
Thank you

PR #331, opened 2018-10-31
Add advisory for symfony/polyfill

paragonie-scott:


BackEndTea:
I think this should also contain `symfony/polyfill-php55`

paragonie-scott:
Should I create two YAML files for this? Or is there a better way to manage that?

nicolas-grekas:
Two yaml please yes.

PR #337, opened 2018-11-21
eZ Systems EZSA-2018-00(7-9)

glye:
Update for EZSA-2018-007, EZSA-2018-008, EZSA-2018-009


ezsystems/ezplatform/2018-11-21-1.yaml
```diff
@@ -0,0 +1,16 @@
+title:     EZSA-2018-009 Do not interpret PHP/PHAR uploads
+link:      http://share.ez.no/community-project/security-advisories/ezsa-2018-009-do-not-interpret-php-phar-uploads
+reference: composer://ezsystems/ezplatform
+branches:
+    2.3.x:
+        time:     2018-11-21 14:08:00
+        versions: ['>=2.3.0','<2.3.2.1']
+    2.2.x:
+        time:     2018-11-21 14:39:00
+        versions: ['>=2.2.0','<2.2.3.1']
```
stof: are 2.0 and 2.1 safe from it ? And other 1.x minor versions ?

If no, I would change constraints to say `<1.7.8.1`, `>=1.8.0,<1.13.4.1` and `>=2.0.0,<2.2.3.1` (another option is to add all other minor versions with ranges like `>=2.0,<2.1` for unpatched versions, but that makes the file a lot bigger for the same result).

stof: btw, this also applies to other advisories here.

glye: Added.

PR #338, opened 2018-11-22
XXE Vulnerability in PHPExcel

MarkBaker:
XXE Vulnerability in PHPExcel


ocramleznem:
Has phpexcel to be mentioned, too?

MarkBaker:
Yes, the same security issue applies equally to both libraries.... both have been patched at the revisions/dates indicated in the respective pull requests

PR #341, opened 2018-12-06
Magento's releases from 2018-11-28

jeroenvermeulen:
SUPEE-10975, v2.1.16 and v2.2.7

### DIFF
--- a/magento/magento1ce/2018-11-28.yaml
+++ b/magento/magento1ce/2018-11-28.yaml
@@ -4,6 +4,6 @@ cve:       ~
 branches:
     1.x:
         time:     2018-11-28 00:00:00
-        versions: ['<1.9.4.0']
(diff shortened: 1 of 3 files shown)


magento/magento1ce/2018-11-28.yaml
```diff
@@ -0,0 +1,9 @@
+title:     SUPEE-10975 - Security enhancements that help close RCE,XSS,CSRF and other vulnerabilities
+link:      https://magento.com/security/patches/supee-10975
+cve:       ~
+branches:
+    1.x:
+        time:     2018-11-28 00:00:00
```
fabpot: it should be the exact time of the merge commit that fixed the issue.

jeroenvermeulen: I updated the times.

fabpot:
Thank you

PR #345, opened 2018-12-11
[FEATURE] Add TYPO3-CORE-SA-2018-005 to TYPO3-CORE-SA-2018-012

pixelbrackets:
Add latest TYPO3 CMS Security Advisories

* https://typo3.org/security/advisory/typo3-core-sa-2018-005/
* https://typo3.org/security/advisory/typo3-core-sa-2018-006/
* https://typo3.org/security/advisory/typo3-core-sa-2018-007/
* https://typo3.org/security/advisory/typo3-core-sa-2018-008/
* https://typo3.org/security/advisory/typo3-core-sa-2018-009/
* https://typo3.org/security/advisory/typo3-core-sa-2018-010/
* https://typo3.org/security/advisory/typo3-core-sa-2018-011/
* https://typo3.org/security/advisory/typo3-core-sa-2018-012/

Package "typo3/cms" addresses legacy packages, "typo3/cms-core"
addresses the sub-tree split packages since v8.x, which became
a mandatory package in v9.0.0.

PR #347, opened 2018-12-22
[FIX] Update security advisories for TYPO3's December 2019 releases

ohader:


typo3/cms-core/2018-12-11-1.yaml
```diff
@@ -5,6 +5,6 @@ branches:
         time:     2018-12-11 14:13:00
         versions: ['>=8.5.0', '<8.7.21']
     9.x:
-        time:     2018-07-12 14:13:00
+        time:     2018-12-11 14:13:00
```
fabpot: Is it the date of the commit fixing the issue?

fabpot: The date here is not the release date but the date (including the time) of the commit that fixes the issue.

ohader: Well, the information that this has been fixed in July 2018 (current state) is just wrong.
That has been introduced with https://github.com/FriendsOfPHP/security-advisories/pull/345/files

With this change at least the date (day) is correct - TYPO3 releases are always released the same day when security fixes are applied to the Git repository (I know that, because I usually run the release process).

Two possibilities on how to continue:
* use the correct date (day) like provided in this PR
* adjust all previous references to point to their according commit date (`CommitDate` not `AuthorDate`)

pixelbrackets: Snap. Yes, Copy & Paste Error -> the date of the commit and the date of the release should have been `2018-12-11` for all patched versions.

pixelbrackets: Ah, TYPO3 now has a [security advisory generator included](https://github.com/TYPO3/darth/commit/4524eec445d4a785127232fea0324e8d4f5cf7f2) into the release process. Nice. This supersedes manual changes like mine and is the reason that the order of the files in this PR changed.

PR #353, opened 2018-12-31
remove optional CVE-key when empty

ocramleznem:
This pr removes all empty CVE-elements. Since this key is optional, it is superfluous to keep empty elements.

### DIFF
--- a/3f/pygmentize/2017-05-15.yaml
+++ b/3f/pygmentize/2017-05-15.yaml
@@ -1,6 +1,5 @@
 title: Remote Code Execution
 link: https://github.com/dedalozzo/pygmentize/issues/1
-cve: ~
 branches:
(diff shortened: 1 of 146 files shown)


fabpot:
I'd like to keep the `cve` entry everywhere as it makes it clear that there is no assigned CVE.

ocramleznem:
For me it is not 100% clear if a `~` marks an not existing CVE or an not-yet-assigned CVE. The CVE can be assigned every tome later. But I understand your desire.

For new issues I will add an empty `cve`-Key if there is any assumption that no CVE will be assigned.

fabpot:
"not existing CVE or an not-yet-assigned CVE" -> that does not a difference to me. Let's keep the current way.

Issue #355, opened 2019-01-10
Issue with security info for root package ezsystems/ezplatform


andrerom:
After addtion of https://github.com/FriendsOfPHP/security-advisories/blob/master/ezsystems/ezplatform/2018-11-21-1.yaml by @glye, many users reports being unable to install latests or update to versions due to our usage of `roave/security-advisories`. In these cases composer is unable to determine version of root package so it gives it a version of `v1.0.0` and composer stops with package conflict.

We have changed `roave/security-advisories` in favour of `sensiolabs/security-checker` in https://github.com/ezsystems/ezplatform/commit/4e94c9552d28961e4251252b1cccd453ac96924f, however that won't help users already with a install. And afaik `sensiolabs/security-checker` does not check root package anyway.

As `roave/security-advisories` auto generates their rules from the source here and thus won't accept changes, I'm wondering if we can add a syntex match for 1.0.x and say =<1.0.0 is valid to avoid it, with some inline comment here about why. Alternative is that we remove it for this package.

Any suggestions/conventions?

Ocramius:
You can say `>=1.0.0`?

Ocramius:
Can you make a set of steps to reproduce the issue?

xabbuh:
Does the entry make any sense at all? I mean will users ever add `ezsystems/ezplatform` as a dependency in their project? IIUC this package is just a template for bootstrapping new applications, isn't it (similar to what `symfony/framework-standard-edition` used to be for Symfony in the past)?

glye:
It looks like the most sensible and easiest solution is to remove the ezplatform entries from security-advisories. I'll do a PR.

Ocramius:
This can only be done if `ezsystems/ezplatform` is not the actual source of the security issue.

andrerom:
Well we could argue `symfony/symfony-standard` is the source, as it's upstream of the repo in question here. But that highly depends on how the fix is done to the issue.

Or we can argue that root repos should not be listed in `roave/security-advisories`.

Ocramius:
> root/project repos should not be listed

If they contain the offending code, they should be listed.

stof:
Well, a bootstrapping template does not really make sense for the advisory repo. A project will never depend on it.

andrerom:
> If they contain the offending code, they should be listed.

There is actually no offending code in this security advisor, it's a enhancement and not strictly a bug:
[...]

PR #356, opened 2019-01-11
Remove ezplatform root repo, as composer can't determine version of root package.

glye:
See https://github.com/FriendsOfPHP/security-advisories/issues/355


stof:
plus, as soon as you use it to create your own project, its version is not the ezplatform one anymore.

PR #358, opened 2019-01-15
api-platform/core: Access control bypass in the GraphQL mutations

dunglas:
Fixed in 2.3.6


xabbuh:
@dunglas Are versions before 2.2.0 affected too?

dunglas:
@xabbuh no, GraphQL support has been introduced in 2.2 (and still is experimental)

garas:
`2019-15-01` date should be `2019-01-15`

PR #364, opened 2019-02-19
Add CVE-2019-5715 in silverstripe/framework

robbieaverill:


robbieaverill:
I've removed the symbiote/silverstripe-multivaluefield report, since it's actually not a security vulnerability as such but broken functionality caused by the fix in silverstripe/framework

fabpot:
Thank you

silverstripe/framework/CVE-2019-5715.yaml
```diff
@@ -0,0 +1,23 @@
+title:     'CVE-2019-5715: Reflected SQL Injection through Form and DataObject'
+link:      https://www.silverstripe.org/download/security-releases/ss-2018-021
+cve:       CVE-2019-5715
+branches:
+    3.6.x:
+        time:     2019-02-19 11:55:05
+        versions: ['>=3.6.0', '<3.6.7']
```
stof: this should be `>=3.0.0` instead, as it says `all versions of SilverStripe 3 prior to 3.6.7, 3.7.3`

robbieaverill: Ok, I got confused by the branch name - will update anyway

robbieaverill: See #367

PR #370, opened 2019-02-21
Fix date for CVE-2019-6338 and CVE-2019-6339

eiriksm:
Just noticed that the year was wrong on the last Drupal CVE files. Probably does not matter that much, and very quick to do so, right after new years :)

Anyway, here is a PR fixing that up.


fabpot:
It does matter actually, thank you.

PR #371, opened 2019-02-22
Add initial dump of drupal contrib SA

### LINKED ISSUE #366: Advisories for Drupal contrib
Hello!

Just wondered if having advisories for Drupal contributed modules in here would be possible/wanted?

They generally do not have a CVE, and of course they would have to be maintained by someone. My thinking was to create a script that parses the advisories available, and programatically create PRs for the ones that were missing. This would help to keep it up to date as well, I could run the script periodically.

Thanks for this package btw, it is very useful!

eiriksm:
Here is an initial dump of Drupal contrib SA.

This fixes #366

I will clean up the script and publish as its own when I get the time.

Edit: Just to clarify: I made the script to parse the latest advisories published. It could theoretically go back in history more, but the format changed some time in 2017. So instead of trying to add parse exceptions, I am going to assume that people with projects not updated for a couple of years, in spite advisories, would not use this repo and its tooling anyway. Open to comments and suggestions on that :)

### DIFF
--- a/drupal/acquia_connector/sa-contrib-2019-014.yaml
+++ b/drupal/acquia_connector/sa-contrib-2019-014.yaml
@@ -1,6 +1,8 @@
 link: 'https://www.drupal.org/sa-contrib-2019-014'
 title: 'Acquia Connector - Moderately critical - Access bypass - SA-CONTRIB-2019-014'
 branches:
-    1.16.x: { time: '2019-02-06 16:38:32', versions: ['>=1.0.0', '<1.16.0'] }
+    1.16.x:
+        time: '2019-02-06 17:13:22'
+        versions: ['>=1.0.0', '<1.16.0']
 composer-repository: 'https://packages.drupal.org/8'
 reference: 'composer://drupal/acquia_connector'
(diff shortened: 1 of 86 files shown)


drupal/acquia_connector/sa-contrib-2019-014.yaml
```diff
@@ -0,0 +1,6 @@
+link: 'https://www.drupal.org/sa-contrib-2019-014'
+title: 'Acquia Connector - Moderately critical - Access bypass - SA-CONTRIB-2019-014'
+branches:
+    1.16.x: { time: '2019-02-06 16:38:32', versions: ['>=1.0.0', '<1.16.0'] }
```
stof: please dump the branch details in expanded format, as done for formatting the advisories edited manually (meaning, increase the depth of the expanded dumping by 1).

If you use the symfony/yaml dumper, that's controlled by the second argument of the method.

eiriksm: 👍 done

vijaycs85:
Probably we should revert back this PR considering we got https://github.com/drupal-composer/drupal-security-advisories for this which has more details of drupal packages and causing issues like https://github.com/Roave/SecurityAdvisories/issues/54

colinodell:
I think this PR is okay because it explicitly defines which repository the packages come from, which Roave/SecurityAdvisories#54 is going to key off of.

vijaycs85:
Thanks @colinodell.  thought it's an overlap as https://github.com/drupal-composer/drupal-security-advisories reading these details from drupal.org. I will follow progress at https://github.com/Roave/SecurityAdvisories/issues/54

fabpot:
reverted

PR #382, opened 2019-03-28
Add ZF2019-01 (zend-developer-tools 1.2.2)

webimpress:
See: https://framework.zend.com/security/advisory/ZF2019-01


webimpress:
Please hold on a minute with merge, as validate gave me an error:
```
Version constraint "1.2.2" is not in an acceptable format.
"versions" must have an upper bound for branch "master".
```

Should I do `['>=1.2.2', '<1.2.3']` instead? Or maybe `['=1.2.2']` is allowed?

Ocramius:
I don't think that downstream tooling can deal with `=1.2.2` either, so your latest commit should be alright.

PR #385, opened 2019-04-07
Remove memory limit in validator usage example

aidantwoods:
Since (at least for me) the validator exceeds PHP's [default memory limit of 128MB](https://www.php.net/manual/en/ini.core.php#ini.sect.resource-limits) when running the validator locally (seems to require about 150MB).

This note might have saved me a couple mins while I thought I somehow caused the validator to break :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,10 +77,9 @@ Contributing security advisories is as easy as it can get:
 
   * If you have a CVE identifier, add it under the `cve` key.
 
-  * Make sure your file validates by running `php validator.php` from the root of this project.
+  * Make sure your file validates by running `php -d memory_limit=-1 validator.php` from the root of this project.
     This script needs some dependencies to be installed via composer, so you need to
-    run `composer install` before. The validator uses a lot of memory, so you might need to
-    use `-d memory_limit=...` (e.g. `php -d memory_limit=200M validator.php`).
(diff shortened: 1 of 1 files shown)


stof:
Yeah, validating that the package exists in the composer repository requires loading some packagist metadata, and 128MB is clearly low for that.

bradjones1:
Not sure this needs a specific call-out though?

README.md
```diff
@@ -79,7 +79,8 @@ Contributing security advisories is as easy as it can get:
 
   * Make sure your file validates by running `php validator.php` from the root of this project.
     This script needs some dependencies to be installed via composer, so you need to
-    run `composer install` before.
+    run `composer install` before. The validator uses a lot of memory, so you might need to
+    use `-d memory_limit=...` (e.g. `php -d memory_limit=200M validator.php`).
```
fabpot: What about `memory_limit=-1`?

aidantwoods: Perhaps just
```diff
-   * Make sure your file validates by running `php validator.php` from the root of this project.
+   * Make sure your file validates by running `php -d memory_limit=-1 validator.php` from the root of this project.
```

in that case? (skip the qualifying statement about using a lot of memory)

fabpot: Works for me

PR #390, opened 2019-04-18
add recent Symfony security advisories

xabbuh:


BackEndTea:
Are versions before 2.7 affected as well? Or are those not included in this repository?

xabbuh:
I am not sure if we need an entry for CVE-2019-10910 in the ProxyManagerBridge too.

symfony/cache/CVE-2019-10912.yaml
```diff
@@ -0,0 +1,32 @@
+title:     "CVE-2019-10912: Prevent destructors with side-effects from being unserialized"
+link:      https://symfony.com/cve-2019-10912
+cve:       CVE-2019-10912
+branches:
+    2.8.x:
```
michaelcullum: This CVE doesn't affect 2.8

symfony/dependency-injection/CVE-2019-10910.yaml
```diff
@@ -0,0 +1,35 @@
+title:     "CVE-2019-10910: Check service IDs are valid"
+link:      https://symfony.com/cve-2019-10910
+cve:       CVE-2019-10910
+branches:
+    2.7.x:
+        time:     2019-04-16 12:48:27
+        versions: ['>=2.7.0', '<2.7.51']
```
stof: The `>=2.7.0` should be removed here, so that the range covers all older versions too, which are also affected

symfony/dependency-injection/CVE-2019-10910.yaml
```diff
@@ -0,0 +1,35 @@
...
+    2.8.x:
+        time:     2019-04-16 12:48:27
+        versions: ['>=2.8.0', '<2.8.50']
+    3.0.x:
+        time:     2019-04-16 12:48:27
```
stof: time should be null, as that branch is not patched (same for all other unpatched branches)

symfony/dependency-injection/CVE-2019-10910.yaml
fabpot: No, the range is only for the 2.7 branch, so removing the lower constraint is not a good idea.

stof: but then, we need to have ranges for all older affected versions too

nicolas-grekas:
> entry for CVE-2019-10910 in the ProxyManagerBridge too.

I think we can, better be safe, the is little to no drawbacks I think.

symfony/security/CVE-2019-10911.yaml
```diff
@@ -0,0 +1,35 @@
...
+    4.0.x:
+        time:     ~
+        versions: ['>=4.0.0', '<4.1.0']
+    4.1.x:
+        time:     2019-04-16 10:54:35
+        versions: ['>=4.1.0', '<4.1.12']
+    4.2.x:
+        time:     2019-04-16 10:54:35
+        versions: ['>=4.2.0', '<4.2.7']
+reference: composer://symfony/securtiy
```
michaelcullum: Typo

michaelcullum:
Thanks @xabbuh

symfony/cache/CVE-2019-10912.yaml
```diff
@@ -0,0 +1,26 @@
+title:     "CVE-2019-10912: Prevent destructors with side-effects from being unserialized"
+link:      https://symfony.com/cve-2019-10912
+cve:       CVE-2019-10912
+branches:
+    3.1.x:
+        time:     ~
+        versions: ['>=3.2.0', '<3.2.0']
```
stof: this one looks wrong

xabbuh: see #403

PR #406, opened 2019-06-26
Magento 2.3.2, 2.2.9 and 2.1.18, 1.x SUPEE-11155 updates

bradjones1:
How in the world did a Drupal Commerce guy end up being the one to submit all the Magento CVEs? Dunno, but here they are! :-)

### DIFF
--- a/magento/magento1ce/2019-06-25.yaml
+++ b/magento/magento1ce/2019-06-25.yaml
@@ -1,0 +1,9 @@
+title:     SUPEE-11155 - RCE, XSS, CSRF and other vulnerabilities
+link:      https://magento.com/security/patches/supee-11155
+cve:       CVE-2019-7893
+branches:
+    1.x:
+        time:     2019-06-25 00:00:00
+        versions: ['>=1.5.0.0', '<1.9.4.2']
+reference: composer://magento/magento1ce
+composer-repository: false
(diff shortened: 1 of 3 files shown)


bradjones1:
Per the Travis failure, the issue with these Magento releases is that they package up something like many dozens of CVE's into one. Should I remove the CVE I did list (just in order to put one in there), then? How do we handle that sort of thing? Or can I list more than one?

fabpot:
@bradjones1 We have a file name convention that should be applied. Yes, we need one file per CVE. If a release contains more than one fix, you can duplicate the file. The title should be a summary of the vulnerability as this is what is going to be displayed to the end user.

bradjones1:
@fabpot Thanks for the clarification. For these framework CVEs, is it general convention that someone at the framework has been a regular contact for adding these? I don't do much Magento but I'm tasked with doing operations for hosting clients, hence my interest here. I know the Drupal ones get added relatively quickly so curious how that's been working.

fabpot:
Most of the time, the CVEs are available before the fix is public. So, it's pretty rare that we update an advisory here because a CVE has been added.

FreekVandeursen:
I created a little script to generate individual files for each CVE. See https://github.com/FriendsOfPHP/security-advisories/pull/416

This pull request can be closed then I guess

PR #415, opened 2019-10-09
Magento 2.2.10 and 2.3.3 release

FreekVandeursen:


FreekVandeursen:
As a sidenote: All security patches for 2.3.3 are also released in 2.3.2-p1, but the validator won't let me add a version number with a letter in it.

magento/magento1ce/CVE-2019-8114.yaml
```diff
@@ -0,0 +1,12 @@
+title: 'PRODSECBUG-2462: Remote code execution via file upload in admin import feature'
+link: https://magento.com/security/patches/magento-2.3.3-and-2.2.10-security-update
+cve: CVE-2019-8114
+reference: composer://magento/magento1ce
+composer-repository: false
+branches:
+  1:
+    time: "2019-10-08 00:00:00"
+    versions:
+    - '>=1'
```
stof: For consistency with the formatting of advisories in this repo, I suggest using the inline syntax for the array of versions.

Note that if you use a script based on symfony/yaml to generate these files, this can be achieved by using the `$inline` parameter of `Yaml::dump()` with a value of `3`. If you use a script using a different Yaml dumper, it may have a similar configuration but I cannot help.

FreekVandeursen: Makes sense @stof , didn't notice the difference with other advisories. I've updated the pull request.

stof:
@FreekVandeursen are you scripting the import of Magento advisories ? If yes, that's a good news, but I would suggest reordering the keys to follow what is most common in this repo (should be easy to apply if a script is generating them, and will be OK for next ones):

- title
- link
- cve
- branches
- reference
- composer-repository

FreekVandeursen:
Yes I created a little script that parses the content of the html page where Magento published the CVE's. So reordering the keys is no problem at all

erikhansen:
@FreekVandeursen Do you have plans to update this repo to support Magento [2.3.2-p2](https://github.com/magento/magento2/releases/tag/2.3.2-p2)? I realize you said that _"All security patches for 2.3.3 are also released in 2.3.2-p1, but the validator won't let me add a version number with a letter in it."_, however it seems important that Magento 2.3.2-p2 can be installed.

I just upgraded a Magento 2.3.2 site to 2.3.2-p2 and I could not install this package:

```
$ composer require magento/product-community-edition=2.3.2-p2 --no-update
$ composer update
$ composer require --dev roave/security-advisories:dev-master
./composer.json has been updated
Gathering patches for root package.
Loading composer repositories with package information
Updating dependencies (including require-dev)
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - roave/security-advisories dev-master conflicts with magento/product-community-edition[2.3.2-p2].
    - roave/security-advisories dev-master conflicts with magento/product-community-edition[2.3.2-p2].
    - roave/security-advisories dev-master conflicts with magento/product-community-edition[2.3.2-p2].
    - roave/security-advisories dev-master conflicts with magento/product-community-edition[2.3.2-p2].
    - Installation request for roave/security-advisories dev-master -> satisfiable by roave/security-advisories[dev-master].
    - Installation request for magento/product-community-edition 2.3.2-p2 -> satisfiable by magento/product-community-edition[2.3.2-p2].
```

Ocramius:
What does `-p2` actually mean? Can't you upgrade to `2.3.3`?

erikhansen:
@Ocramius More details about security-only patch releases here: https://community.magento.com/t5/Magento-DevBlog/Introducing-the-New-Security-only-Patch-Release/ba-p/141287

2.3.3 includes many feature/bug fixes. Historically this means that the upgrade takes a long time, and many merchants opt not to upgrade right away. The security-only patch release allows a merchant to only get the security fixes.

Ocramius:
In order to support things like `2.3.2-p2`, we'd need a specification for sorting these suffixes. Most certainly possible to fetch it from composer's SAT internals.

erikhansen:
@Ocramius Magento is going to be adopting this versioning nomenclature going forward, so hopefully it won't be too hard to figure out how to support it.

hardyjohnson:
Could you use an inline alias in the composer file to get the necessary sorting (if Magento won't adopt the four dot version notation)?
https://getcomposer.org/doc/articles/aliases.md#require-inline-alias

```
  "require": {
    "magento/product-enterprise-edition": "2.3.2-p1 as 2.3.2.1",

  "require": {
    "magento/product-enterprise-edition": "2.3.2-p2 as 2.3.2.2",
```
etc.

FreekVandeursen:
@erikhansen As far as I know 2.3.2-p1 is a perfectly valid composer version number. See https://getcomposer.org/doc/04-schema.md#version where it is even listed specifically as an example: v2.0.4-p1

But to be able to use such version numbers here, at least the yaml validator needs to be updated to support such version numbers, and optionally the security checker as well. But since this versioning system is explicitly allowed by composer I think it would make sense to do so.

If such versioning can be allowed here I'll be happy to update the correct versions to the Magento packages here.

Any ideas @stof or @fabpot ?

Ocramius:
@FreekVandeursen @erikhansen @hardyjohnson it seems like downstream `-p(atch)?(\d*)` is already supported thanks to @slash3b's work: https://github.com/Roave/SecurityAdvisoriesBuilder/blob/95f1c1039f69bc5f045c5e3a7075eb79a7d1afc5/src/Roave/SecurityAdvisories/Version.php#L40-L41

Need to adapt it in here, and then you can update the definitions :+1:

Ocramius:
Then the linked code needs fixing, but indeed, `-p*` is confusing

FreekVandeursen:
I have an open pull request to fix this @kassner , but until it is accepted I cannot push the 2.3.2-p2 as acceptable version. Unfortunately I haven't gotten a reply yet whether my fix can be accepted or if there needs to be fixed more than just this regex in the validator.

In the meantime, did you try the alias workaround as mentioned above by @erikhansen ?

kassner:
@FreekVandeursen with the generated advisories over https://github.com/Roave/SecurityAdvisories having this, it would work.

```
"magento/product-community-edition": ">=2,<2.2.10|>=2.3,<=2.3.2,2.3.2-p1",
```

Can't we achieve that with current code?

kassner:
The alias doesn't work, as composer compares the original version:

```
"magento/product-enterprise-edition": "2.3.2-p2 as 2.3.3",
```

```
Dependency resolution completed in 0.046 seconds
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - magento/product-community-edition 2.3.2-p2 conflicts with roave/security-advisories[dev-master].
    - roave/security-advisories dev-master conflicts with magento/product-community-edition[2.3.2-p2].
    - magento/product-enterprise-edition 2.3.2-p2 requires magento/product-community-edition 2.3.2-p2 -> satisfiable by magento/product-community-edition[2.3.2-p2].
    - Installation request for magento/product-enterprise-edition 2.3.2-p2 as 2.3.3 -> satisfiable by magento/product-enterprise-edition[2.3.2-p2].
    - Installation request for roave/security-advisories dev-master -> satisfiable by roave/security-advisories[dev-master].
```

FreekVandeursen:
Unfortunately @kassner, this repository has a validator function which validates the version numbers in the security advisories. And for now, that validator doesn't recognize version numbers with -p or -patch in it.

So until @fabpot  accepts pull request https://github.com/FriendsOfPHP/security-advisories/pull/417 I cannot update the advisories here.
There might actually be good reasons for not accepting that pull request, as I'm not sure if the symfony security check can handle the -p / -patch version numbers. In that case removing the magento security advisories at all would make more sense I guess, since having false positives is a bad situation too.

kassner:
@FreekVandeursen thanks for stepping up and sending the PR, much appreciated!

PR #417, opened 2019-10-11
Allow version numbers with patch number

FreekVandeursen:


FreekVandeursen:
@fabpot Do you know if the symfony security check will work with version numbers with -p or -patch ? If not, there's no need for this pull request anyway.

PR #418, opened 2019-10-16
Corrects the vulnerable version of PHP Excel

moebrowne:
The PHP Excel project released another version which basically archived the project but importantly didn't fix this vulnerability.

Changes in the 1.8.2 release: https://github.com/PHPOffice/PHPExcel/compare/1.8.2...1.8

### DIFF
--- a/phpoffice/phpexcel/CVE-2018-19277.yaml
+++ b/phpoffice/phpexcel/CVE-2018-19277.yaml
@@ -4,5 +4,5 @@ cve:       CVE-2018-19277
 branches:
     master:
         time:       2018-11-22 23:07:00
-        versions:   ['<=1.8.1']
+        versions:   ['<=1.8.2']
 reference: composer://phpoffice/phpexcel


BackEndTea:
How about setting an uperbound of version `99.99.99` or something like that, so any new tags won't fall outside of this range?

Idk what the standard is for these cases

moebrowne:
In that case shouldn't the version constraint be `*`?

fabpot:
I think I'd prefer the 9.9.99 trick

stof:
`*` is currently not accepted by the validator.

phpoffice/phpexcel/CVE-2018-19277.yaml
```diff
@@ -4,5 +4,5 @@ cve:       CVE-2018-19277
 branches:
     master:
         time:       2018-11-22 23:07:00
```
stof: time should be null, as there is no fix

PR #424, opened 2019-11-19
Add Magento version 2.3.2-p2

FreekVandeursen:
Validation depending on https://github.com/FriendsOfPHP/security-advisories/pull/417

### DIFF
--- a/magento/product-community-edition/CVE-2019-8090.yaml
+++ b/magento/product-community-edition/CVE-2019-8090.yaml
@@ -7,6 +7,6 @@ branches:
         versions: ['>=2.2', '<2.2.10']
     '2.3':
         time: '2019-10-08 00:00:00'
-        versions: ['>=2.3', '<2.3.3']
+        versions: ['>=2.3', '<2.3.3', '<2.3.2-p2']
 reference: 'composer://magento/product-community-edition'
 composer-repository: false
--- a/magento/product-community-edition/CVE-2019-8092.yaml
+++ b/magento/product-community-edition/CVE-2019-8092.yaml
@@ -7,6 +7,6 @@ branches:
         versions: ['>=2.2', '<2.2.10']
     '2.3':
         time: '2019-10-08 00:00:00'
-        versions: ['>=2.3', '<2.3.3']
+        versions: ['>=2.3', '<2.3.3', '<2.3.2-p2']
 reference: 'composer://magento/product-community-edition'
 composer-repository: false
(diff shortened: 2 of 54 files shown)


Ocramius:
@FreekVandeursen this produces some bizarre effects downstream: https://github.com/Roave/SecurityAdvisories/commit/b04afa5560ac7f34fe6a24c620ccb199d61624a0

Also interesting for @NikolaPosa

FreekVandeursen:
Oh wow, didn't see that coming @Ocramius . I thought this was the correct syntax to specify this condition. Or do you think I made a mistake in the yaml files?

Ocramius:
No, I think it's fine in this repository, but as you can see, downstream does not handle `-p\d+` very well

nikolaposa:
Thanks for the heads up Marco. When I have time I'll investigate how https://github.com/nikolaposa/version copes with this syntax. 

I'm noticing that FAQ section on the official semver page has changed a lot: https://semver.org/. I like this part: https://semver.org/#is-v123-a-semantic-version. Also like that someone has contributed regex: https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string. I might incorporate it into my library.

FreekVandeursen:
Are you sure the -p is the problem @Ocramius ? Isn't it the 3 version numbers in a single constraint?

stof:
The 2 upper bounds indeed look like the issue, as that's not defining a range anymore:

- if both need to be satisfied at the same time, `<2.3.3` is useless here
- if you consider it as a OR, `<2.3.2-p2` is useless here (as 2.3.2-p2 would still match `<2.3.3` and so be marked as affected)

FreekVandeursen:
My intention was @stof : 2.3.2 is vulnerable. 2.3.2-p2 and 2.3.3 are safe. Will [">2.3", "<2.3.2-p2"] be sufficient then?

stof:
well, yes, because `2.3.3` is out of the range `>= 2.3, <2.3.2-p2`

FreekVandeursen:
Unfortunately that will not be enough. When I change that, the SecurityAdvisoriesBuilder will translate [">2.3", "<2.3.2-p2"] into "magento/product-community-edition": ">=2.3,<2.3.2". That is not correct since 2.3.2 is vulnerable

fabpot:
What is `SecurityAdvisoriesBuilder`? This database should not care about tools. Tools should be adapted when needed.

FreekVandeursen:
You're right @fabpot , I should have put that comment in another repository. I'll update the database here to match the constraints >2.3, <2.3.2-p2

stof:
Well, the https://github.com/Roave/SecurityAdvisoriesBuilder repo already has an issue about the fact that it does not handle 2.3.2-p2 properly. But AFAICT, it won't change the upper bound. It will simply fail to optimize the version range.

PR #426, opened 2019-11-19
Add a validator ensuring that each branch has a well-formed range

stof:
It does not make sense to have multiple upper bounds for a range, as they would need to match all, meaning that the lowest version of them would be the useful one. Same is true for lower bounds (except that the highest version is the meaningful one).
Defining multiple affected ranges is what branches are representing in the advisory format.

Build will be green once #425 is merged

### DIFF
--- a/validator.php
+++ b/validator.php
@@ -179,10 +179,20 @@ final class Validate extends Command
                             }
 
                             if ('<' === substr($version, 0, 1)) {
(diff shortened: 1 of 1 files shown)


stof:
Travis build before merging #425, showing that it detects that case: https://travis-ci.org/FriendsOfPHP/security-advisories/builds/614028742

PR #430, opened 2019-12-04
Add CVE-2019-3465 for SimpleSAMLphp

artfreyr:
I have added this CVE for SimpleSAMLphp because it is not being detected when scanned

### DIFF
--- a/simplesamlphp/saml2/CVE-2019-3465.yaml
+++ b/simplesamlphp/saml2/CVE-2019-3465.yaml
@@ -1,0 +1,14 @@
+title:     Signature validation bypass
+link:      https://simplesamlphp.org/security/201911-01
+cve:       CVE-2019-3465
+branches:
+    2.x:
+        time:     2019-11-06 13:20:00
+        versions: ['>=2.0.0', '<2.3.9']
+    3.x:
+        time:     2019-11-06 13:20:00
+        versions: ['>=3.0.0', '<3.4.2']
+    4.x:
+        time:     2019-11-06 13:20:00
+        versions: ['>=4.0.0', '<4.1.1']
+reference: composer://simplesamlphp/saml2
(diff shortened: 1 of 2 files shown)


simplesamlphp/saml2/CVE-2019-3465.yaml
```diff
@@ -0,0 +1,14 @@
...
+    2.x:
+        time:     2019-11-06 13:20:00
+        versions: ['>=2.0.0', '<2.3.9']
+    3.x:
+        time:     2019-11-06 13:20:00
+        versions: ['>=3.0.0', '<3.4.2']
+    4.x:
+        time:     2019-11-06 13:20:00
+        versions: ['>=4.0.0', '<4.1.1']
+reference: composer://simplesamlphp/saml2
```
stof: Is this advisory actually needed ? The actual CVE is in `robrichards/xmlseclibs`, which is a dependency of `simplesamlphp/saml2`.
And it is actually possible to have an older version of `simplesamlphp/saml2` installed with a patched version of `robrichards/xmlseclibs`. These "fixed" version of `simplesamlphp/saml2` are the one where they bumped the min requirement for `robrichards/xmlseclibs`, but that does not make older versions of  `simplesamlphp/saml2` vulnerable by themselves (we need an advisory covering `robrichards/xmlseclibs` though, for things to be detected)

artfreyr: Hey @stof, thank you for your response. I ran the test on the "affected" version of SimpleSAMLphp I installed today and did not get a detection. But you are right that the vulnerability lies in in `robrichards/xmlseclibs`. And it explains why I did not get a detection because I was no longer using an affected version of `robrichards/xmlseclibs`.

There already exists an advisory for `robrichards/xmlseclibs` for CVE-2019-3465, so I shall close this pull request.

PR #433, opened 2019-12-20
Add security advisories for scheb/two-factor-bundle

scheb:
Addding security advisories for `scheb/two-factor-bundle`. One recent one and an older one (but I still see people using that version).


scheb:
That was quick, thanks for merging!

PR #447, opened 2020-02-18
Add silverstripe/framework issue CVE-2019-19325

maxime-rainville:


silverstripe/framework/CVE-2019-19325.yaml
```diff
@@ -0,0 +1,11 @@
+title:     'CVE-2019-19325: XSS through non-scalar FormField attributes'
+link:      https://www.silverstripe.org/download/security-releases/cve-2019-19325/
+cve:       CVE-2019-19325
+branches:
+    4.0.x:
+        time:     2020-02-16 23:54:40
+        versions: ['>=4.0.0', '<4.4.5']
```
xabbuh: Is this correct? The blog post says that `^4.3` versions are affected. Thus I would have expected `>=4.3.0` as the lower bound.

maxime-rainville: Version 4.0 to 4.3 have been deprecated. The issue was originally identified on 4.3. It's probably in 4.0 has well, but because we don't support it any more, we didn't go back and test specifically for it.

xabbuh: thanks 👍

PR #451, opened 2020-03-13
Add Laravel possible XSS Attack in blade component

Justin991q:


stof:
The same advisory should also be added for the `illuminate/view` subtree package.

Justin991q:
> The same advisory should also be added for the `illuminate/view` subtree package.

Added

michaelcullum:
@Justin991q There were two pull requests for this same advisory so this has been merged in #453 but thank you for your PR.

PR #459, opened 2020-04-10
add CVE against reveal.js

drsassafras:


fabpot:
This repo is about PHP security issues, looks like a JS lib here with no Packagist entry.

drsassafras:
Did I not get the packagist entry correct? https://packagist.org/packages/components/revealjs

drsassafras:
Or perhaps you are saying that because it is a JS dependency listed on packagist that it is excluded from security advisories here?

garas:
JS packages should be primarily installed with npm/bower/yarn. Composer should be used for PHP libraries.

https://github.com/components has many Composer wrappers for various JS packages and many of them has one or more CVE. Do you want duplicate them all here?

PR #473, opened 2020-07-03
Add nystudio107/craft-seomatic - 2020-04-28-1

brandonkelly:


brandonkelly:
@fabpot The validator is saying `Invalid branch name "v3".`, but that’s the actual branch name (https://github.com/nystudio107/craft-seomatic/tree/v3) so not sure what to do about that.

fabpot:
Should be 3.x IIRC (in my phone right now)

fabpot:
Should be 3.x IIRC (in my phone right now)

brandonkelly:
Thanks, that’s passing locally. Not sure I understand why `3.x` is acceptable but not `v3`, when `v3` is the actual Git branch name. Would be good if the docs could clarify that a bit more.

PR #478, opened 2020-07-28
Add Laravel vuln in cookie session driver

jaylinski:
See https://blog.laravel.com/laravel-cookie-security-releases.

I'm not sure if the 5.x branch should be marked as unfixed... :thinking: 

Also: should there be an additional entry for the `illuminate/cookie` package?

### DIFF
--- a/illuminate/cookie/2020-07-27-1.yaml
+++ b/illuminate/cookie/2020-07-27-1.yaml
@@ -1,0 +1,14 @@
+title:     RCE vulnerability in "cookie" session driver
+link:      https://blog.laravel.com/laravel-cookie-security-releases
+cve:       ~
+branches:
+    "5.5":
+        time:     ~
+        versions: ['>=5.5.0', '<=5.5.44']
+    "6.x":
+        time:     2020-07-28 18:34:00
+        versions: ['>=6.0.0', '<6.18.31']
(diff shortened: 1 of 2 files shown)


devfrey:
Could this be merged please? (ping @fabpot)

stof:
shouldn't this also register the advisory for one of the `illuminate/*` subtree split ?

stof:
I would say that `illuminate/cookie` is the impacted package

stof:
@taylorotwell do you know whether 5.x was also affected by this security issue or whether this issue was introduced as part of 6.0 ? For the advisory, it would make sense to mark 5.x as affected (without any patched version) if it is indeed the case.

jaylinski:
@fabpot I added an entry for `illuminate/cookie`. I also added the (unpatched) `5.5` branch, since it is still supported according to https://laravel.com/docs/7.x/releases#support-policy.

GrahamCampbell:
This vulnerability affects every laravel version since 4.1.0, but this PR doesn't take this into account.

GrahamCampbell:
> For the advisory, it would make sense to mark 5.x as affected (without any patched version) if it is indeed the case.

Yes, Laravel 4 too. I have sent in a PR for this.

PR #479, opened 2020-08-06
Add Laravel vuln in eloquent models

jaylinski:
See https://blog.laravel.com/security-release-laravel-61834-7232.

I'm not sure if the 5.x branch should be marked as unfixed... :thinking: 

Also: should there be an additional entry for the `illuminate/database` package?

### DIFF
--- a/illuminate/database/2020-08-06.yaml
+++ b/illuminate/database/2020-08-06.yaml
@@ -1,0 +1,14 @@
+title:     Guard bypass in Eloquent models
+link:      https://blog.laravel.com/security-release-laravel-61834-7232
+cve:       ~
+branches:
+    "5.5":
+        time:     ~
+        versions: ['>=5.5.0', '<=5.5.44']
+    "6.x":
+        time:     2020-08-06 14:55:00
+        versions: ['>=6.0.0', '<6.18.34']
+    "7.x":
+        time:     2020-08-06 14:56:00
+        versions: ['>=7.0.0', '<7.23.2']
+reference: composer://illuminate/database
--- a/laravel/framework/2020-08-06-1.yaml
+++ b/laravel/framework/2020-08-06-1.yaml
@@ -1,0 +1,14 @@
+title:     Guard bypass in Eloquent models
+link:      https://blog.laravel.com/security-release-laravel-61834-7232
+cve:       ~
+branches:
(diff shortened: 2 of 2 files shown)


fabpot:
You should indeed list `illuminate/database` if it is affected as well.

jaylinski:
@fabpot I added an entry for `illuminate/database`. I also added the (unpatched) `5.5` branch, since it is still supported according to https://laravel.com/docs/7.x/releases#support-policy.

illuminate/database/2020-08-06.yaml
```diff
@@ -0,0 +1,14 @@
+title:     Guard bypass in Eloquent models
+link:      https://blog.laravel.com/security-release-laravel-61834-7232
+cve:       ~
+branches:
+    5.5:
```
fabpot, line 5: Needs to be quoted (`"5.5:"`), same below

stof: actually, `"5.5":`

laravel/framework/2020-08-06-1.yaml
```diff
@@ -0,0 +1,14 @@
+title:     Guard bypass in Eloquent models
+link:      https://blog.laravel.com/security-release-laravel-61834-7232
+cve:       ~
+branches:
+    5.5:
+        time:     ~
+        versions: ['>=5.5.0', '<=5.5.49']
```
stof, line 7: I would put `<60` if this is unpatch, to make it clearer

jaylinski: Are you sure? Laravel did not use semver until v6, so v5.5 is actually a major version.

illuminate/database/2020-08-06.yaml
jaylinski: @fabpot Fixed it.

Note to myself: check CI status for errors when doing a PR

GrahamCampbell:
This vulnerability affects every laravel version since 4.1.0, but this PR doesn't take this into account.

jaylinski:
@GrahamCampbell Since you opened #480, you maybe also want to check the listed versions in this PR.

GrahamCampbell:
Yep. I will do that soon, once I have established exactly which versions are affected.

PR #480, opened 2020-08-14
Laravel's cookie encryption vulerability affects all versions since 4.1.0

GrahamCampbell:


jaylinski:
Since I submitted the recent Laravel advisories:

My understanding was to not include end-of-life versions (https://laravel.com/docs/7.x/releases#support-policy) in these advisories, since it does not make sense to use an unsupported version if you care about security.

But there are probably users out there that run an unsupported Laravel version and still expect to be warned about security issues.

That's a call @fabpot has to make.

GrahamCampbell:
It would be exceptional not to document all affected versions. Every other vulnerability in this package documents **all** vulnerable versions, including the laravel ones. Missing versions should be treated as a bug.

GrahamCampbell:
Take for example, the last cookie vuln in laravel: https://github.com/FriendsOfPHP/security-advisories/blob/master/laravel/framework/2018-08-08-1.yaml.

GrahamCampbell:
Laravel 4 was already dead for a long time, even in 2018, but was still documented.

GrahamCampbell:
Also, if we don't tell users there are vulnerabilities, then how will they know to upgrade. One of the purposes of this security database is to help users running older software know when they need to patch something, or even to encourage them to upgrade to a supported version. ;)

laravel/framework/2020-07-27-1.yaml
```diff
@@ -2,9 +2,39 @@ title:     RCE vulnerability in "cookie" session driver
...
+    "5.2":
+        time:     ~
+        versions: ['>=4.2.0', '<=4.2.99999']
+    "5.0":
+        time:     ~
+        versions: ['>=5.0.0', '<=5.0.99999']
+    "5.1":
+        time:     ~
+        versions: ['>=5.1.0', '<=5.1.99999']
+    "5.2":
```
fabpot, line 17: 5.2 is duplicate.

illuminate/cookie/2020-07-27-1.yaml
```diff
@@ -2,9 +2,39 @@ title:     RCE vulnerability in "cookie" session driver
...
+    "5.2":
+        time:     ~
+        versions: ['>=4.2.0', '<=4.2.99999']
+    "5.0":
+        time:     ~
+        versions: ['>=5.0.0', '<=5.0.99999']
+    "5.1":
+        time:     ~
+        versions: ['>=5.1.0', '<=5.1.99999']
+    "5.2":
```
fabpot, line 17: 5.2 is also duplicated in this file.

GrahamCampbell: Not duplicated, but still wrong. The first occurrence was a typo for 4.2. I'll get that fixed. ;)

PR #489, opened 2020-09-18
Add Drupal SA-CORE-2020-007 - 011

eiriksm:


stof:
@eiriksm according to the link to the Drupal advisory, there is a CVE assigned to it. But it is missing in the YAML here.

azlanroyz11394 (commented):
@eiriksm thank you all my colleagues 😭😭😭 I am weak and not enough experience spilling your help all who I am..thank you very much I can not repay your services all

eiriksm:
@stof Ah right, thanks. I added a PR with the missing CVEs (and some more) in #490

PR #493, opened 2020-10-02
add recent Nette security advisories

dg:


nette/nette/CVE-2020-15227.yaml
```diff
@@ -0,0 +1,11 @@
+title:     "CVE-2020-15227: Potential Remote Code Execution vulnerability"
+link:      https://blog.nette.org/en/cve-2020-15227-potential-remote-code-execution-vulnerability
+cve:       CVE-2020-15227
+branches:
+    2.0.x:
+        time:     2020-08-25 03:50:00
+        versions: ['>=2.0.0', '<2.0.19']
```
stof, line 7: is 2.0.0 the first affected version or is 1.x also subject to the RCE ? If 1.x is also affected, you should remove the lower bound here

stof:
this misses the advisory for `nette/application`, which is the one affected for newer versions

PR #497, opened 2020-10-18
Add CVE-2019-16738

Rudloff:
I noticed there wasn't any MediaWiki CVE here.
I don't have time to add all of them, but I figured I would do my part and at least add one.


legoktm:
I'm a little late here, but what's the value of adding MediaWiki's CVEs to this repository? It's not really installable as a composer package and the fact that it's listed on packagist is mostly a historical accident (it's been marked as abandoned for a while now).

Rudloff:
I think it is still useful to have a standardized way to list vulnerabilities for a PHP project.
And I guess it shows that the latest version on Packagist is vulnerable and should not be used.

(IMHO MediaWiki should be installable with Composer but that's another debate.)

legoktm:
> I think it is still useful to have a standardized way to list vulnerabilities for a PHP project.

My understanding is that this repository is primarily intended for projects published/obtained through composer, not a large software application like MediaWiki in which you really just have to run the latest version. Should vulnerabilities in bundled jQuery also be included here?

To be clear, the main reason I'm asking these questions is because the current list is woefully incomplete and outdated, so if we do want to include MediaWiki vulnerabilities here we should make it part of the release process, otherwise it's not useful at all.

Compare to what the [Debian security tracker](https://security-tracker.debian.org/tracker/source-package/mediawiki) has. Or just the general MediaWiki release notes or CVE feed...what's the value in having it in this database?

>  And I guess it shows that the latest version on Packagist is vulnerable and should not be used.

All of the versions listed on Packagist are vulnerable and outdated and shouldn't be used. But it's not even possible to use it through Packagist in the first place...

In short, I think having a very incomplete list here is detrimental as it does not properly reflect the security status of MediaWiki. So if there's value in this (I still don't see what it is), we should either have all of them included here and a plan for keeping it up to date, or just drop it all.

Rudloff:
> But it's not even possible to use it through Packagist in the first place...

Well, if we ignore the fact that the package is not updated anymore, it is totally possible to install MediaWiki with Composer (see [this proof of concept](https://github.com/Rudloff/mediawiki-composer), for example).
And one of the reasons I added MediaWiki CVEs in this repository is that I have clients that install MediaWiki with Composer (but I now have to maintain [a custom repository](https://mwpackagist.netlib.re/repo/#mediawiki/core) so that they can keep updating it).

Anyway, I understand your concerns and how it would not be possible or useful for the MediaWiki team to systematically publish security advisories here.

PR #498, opened 2020-10-18
Add CVE-2018-16831

Rudloff:
It looks like there is another CVE for the same vulnerability as CVE-2018-13982.


fabpot:
Thank you

PR #499, opened 2020-10-19
Add 2020 CVEs for MediaWiki

Rudloff:
I am not entirely sure about the constraint for vulnerabilities that only affect the RC versions of 1.35 :slightly_frowning_face:


fabpot:
Thank you.

PR #501, opened 2020-11-17
[TASK] Add security advisories for TYPO3's November 2020 releases

ohader:


typo3/cms-core/2020-11-17-1.yaml
```diff
@@ -0,0 +1,13 @@
...
+    10.x:
+        time: '2020-11-17 08:50:35'
+        versions: ['>=10.0.0', '<10.4.10']
+    9.x:
+        time: '2020-11-17 08:49:44'
+        versions: ['>=9.0.0', '<9.5.23']
+    8.x:
+      time: '2020-11-17 08:55:33'
+      versions: ['>=8.7.0', '<8.7.38']
+reference: 'composer://typo3/cms-core'
```
stof, line 13: As the issue happens in `typo3fluid/fluid`, and that `typo3/cms-core` does not depend on an exact version of fluid, this does not guarantee that the advisory matching is right for projects using composer.
Instead, the advisory should be assigned to the `typo3fluid/fluid` package itself

stof (changes requested):

typo3/cms-core/2020-11-17-1.yaml
ohader: `typo3fluid/fluid` is totally separate thing... however `typo3/cms` has strong dependencies on it as well.

→ see Fluid dependency https://github.com/TYPO3/TYPO3.CMS/blob/master/composer.json#L80
→ at least `2.6.10` having the latest security fixes included

stof: Yeah, but as the security issue is in typo3fluid/fluid (at least according to my understanding of the linked page), the advisory should be on that package, not on `typo3/cms` and `typo3/cms-core` which depend on it.
Bundled versions of typo3fluid/fluid in the typo3 CMS full download are not meaningful here, because composer-based installation are not relying on such bundled dependencies but on composer resolving dependencies.

stof: > at least 2.6.10 having the latest security fixes included

According to https://typo3.org/security/advisory/typo3-core-sa-2020-009/, that's not only fixed in 2.6.10 but also in patch releases for older minor versions of typo3fluid/fluid:

> Update to versions 2.0.8, 2.1.7, 2.2.4, 2.3.7, 2.4.4, 2.5.11 or 2.6.10 of the underlying standalone *typo3fluid/fluid* package.

ohader: Okay fine, so let's skip TYPO3-CORE-SA-2020-009 for this PR

typo3/cms-core/2020-11-17-2.yaml
```diff
@@ -0,0 +1,13 @@
...
+    10.x:
+        time: '2020-11-17 08:51:00'
+        versions: ['>=10.0.0', '<10.4.10']
+    9.x:
+        time: '2020-11-17 08:49:54'
+        versions: ['>=9.0.0', '<9.5.23']
+    8.x:
+      time: '2020-11-17 08:55:33'
+      versions: ['>=8.7.0', '<8.7.38']
+reference: 'composer://typo3/cms-core'
```
stof, line 13: this looks wrong to me. the linked page says that this affects `typo3/cms-fluid`, not `typo3/cms-core`

typo3/cms-core/2020-11-17-4.yaml
```diff
@@ -0,0 +1,7 @@
+title: 'TYPO3-CORE-SA-2020-012: XML External Entity in Dashboard Widget'
+link: 'https://typo3.org/security/advisory/typo3-core-sa-2020-012/'
+branches:
+    10.x:
+        time: '2020-11-17 08:51:21'
+        versions: ['>=10.0.0', '<10.4.10']
+reference: 'composer://typo3/cms-core'
```
stof, line 7: should be assigned to `typo3/cms-dashboard`

stof (changes requested):

typo3/cms-core/2020-11-17-1.yaml
stof: you should not skip it entirely. You should create the advisory for the `typo3fluid/fluid` package instead.

typo3/cms-core/2020-11-17-4.yaml
ohader: None of the TYPO3 packages can be used without `typo3/cms-core`, that's the reason we're using it for these advisories. Each of the packages have the same versions since they are maintained in a monolithic repository and subtree-split into those packages. Basically we did it like that during the last few years...

Since none of these `typo3/cms-*` packages can be used standalone, I don't see a reason for making things more complicated here. I'm giving it up for the time being...

stof: @ohader but is it enforced that you have the exact same version of `typo3/cms-core` and `typo3/cms-fluid` ?

IIRC, the generation of these advisories has been scripted by the typo3 team based on the advisory pages. And these pages do contain that info already (see `ext: core` vs `ext: dashboard` in the `Subcomponent` info) so that could be extracted automatically.

ohader: Okay, let's have a quick look into that:
* https://packagist.org/packages/typo3/cms-fluid#v10.4.10 → require `typo3-core:10.4.10` (exact version, no `^`, no `~`)
* https://packagist.org/packages/typo3/cms-core?query=typo3%2Fcms&type=typo3-cms-framework#v10.4.10 → same for all subtree-split packages `typo3/cms-*`
* all(!) packages require a specific version of `typo3/cms-core`

All of those packages have to be used for the exact same version - as the TYPO3 core monolith is maintained and released as a whole.

fabpot:
I'm going to merge this one as is. If there is anything to change or tweak, another PR could be submitted.

PR #502, opened 2020-11-18
Add 2018-2019 CVEs for MediaWiki

Rudloff:


mediawiki/core/CVE-2018-0503.yaml
```diff
@@ -0,0 +1,17 @@
+title:     $wgRateLimits (rate limit / ping limiter) entry for 'user' overrides that for 'newbie'
+link:      https://phabricator.wikimedia.org/T169545
+reference: composer://mediawiki/core
+cve:       CVE-2018-0503
+branches:
+    1.27.x:
+        time: 2018-09-20 18:59:33
+        versions: ['>=1.27.0', '<1.27.5']
```
stof, line 8: Is `1.26` (and older) safe from the issue ? If no, this lower bound should probably be removed, to cover all older versions.

mediawiki/core/CVE-2018-0503.yaml
```diff
@@ -0,0 +1,17 @@
...
+    1.29.x:
+        time: 2018-09-20 19:24:54
+        versions: ['>=1.29.0', '<1.29.3']
```
stof, line 11: Was 1.28 safe ? If no, you either miss an affected range or should change that lower bound.

mediawiki/core/CVE-2018-0503.yaml
Rudloff: 1.26 is probably vulnerable but it was EOL at the time so I don't think it was investigated which exact version introduced the vulnerability.
Should we remove the lower bound even if it means flagging old versions that might not be vulnerable?

mediawiki/core/CVE-2018-0503.yaml
Rudloff: Same as above, 1.28 was EOL at the time.

mediawiki/core/CVE-2018-0503.yaml
stof: Well, I think it is fine if some EOL versions are marked as affected while they were actually not (you don't want people to use EOL versions, and such version will be safe to use only if it is unaffected by **all** newest advisories).
On the other hand, not marking EOL versions as affected might mean that tooling checking things considers that the last 1.26.x release is fine while it is actually affected by many vulnerabilities.

PR #508, opened 2020-11-26
Fix several entries time that breaks time parsing

tucksaun:


stof:
@tucksaun this should be catched by the validator script IMO

tucksaun:
@stof I totally agree but I don't have enough time to take care of it before the end of SymfonyWorld 😅

PR #509, opened 2020-11-26
Add Drupal SA-CORE-2020-013

ericbarrette:


stof:
these 2 CVEs are actually not about the drupal package, but about `pear/archive_tar`, which is a dependency of drupal. So that's the package to which the advisory needs to be attached (which also benefits other packages using it)

ericbarrette:
Good catch, I'm assuming they will be raised in Drupal projects if we do so?

stof:
given that `drupal/core` and `drupal/drupal` depend on `pear/archive_tar` rather than bundling it, yes, because your composer.lock file contains `pear/archive_tar`

stof:
And this was already added in https://github.com/FriendsOfPHP/security-advisories/pull/507 btw

ericbarrette:
Ok thank you, I'm closing this PR.

PR #517, opened 2020-12-28
Configure CI to run the validator

legoktm:
I usually forget to run this, so it would be good if CI did it automatically.

### DIFF
--- a/.github/workflows/php.yaml
+++ b/.github/workflows/php.yaml
@@ -1,0 +1,27 @@
+name: Validation
+
+on:
+  push:
+  pull_request:
+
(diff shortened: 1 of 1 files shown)


fabpot:
That's already the case via the `.travis.yml` file, am I missing something?

legoktm:
> That's already the case via the `.travis.yml` file, am I missing something?

I didn't see any feedback from Travis on #516, where my initial patch failed the validator. Nor do I see any on this patch...

stof:
@fabpot probably broken due to Travis ending free support for open-source projects.

PR #522, opened 2021-01-15
Add mautic and recent CVEs

RCheesley:
I have created a PR according to the instructions to add [Mautic](https://github.com/mautic/mautic) and included three recent CVEs, however I am unable to have it pass the validation due to the naming of our main branch (features).

Would appreciate some guidance on how to proceed!

### DIFF
--- a/mautic/core/CVE-2020-35124.yaml
+++ b/mautic/core/CVE-2020-35124.yaml
@@ -5,13 +5,13 @@ branches:
     features:
         time:     2021-01-14 17:07:00
         versions: ['<=3.2.3']
-    3.2:
+    '3.2':
         time:     2021-01-14 17:07:00
         versions: ['<=3.2.3']
-    3.1:
+    '3.1':
         time:     2021-01-14 17:07:00
         versions: ['<=3.2.3']
-    2.16:
+    '2.16':
         time: 2021-01-14 17:07:00
-        versions: [ '<=2.16.4' ]
+        versions: ['<=2.16.4']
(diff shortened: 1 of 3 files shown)


fabpot:
Just remove the branch, it's not needed (we only track release branches).

RCheesley:
Great, thank you, should be good to go!

mautic/core/CVE-2020-35124.yaml
```diff
@@ -0,0 +1,14 @@
+title: Mautic core - Highly Critical - XSS vulnerability leveraged through referrers could allow un-authorized admin access
+link: https://www.mautic.org/blog/community/security-release-all-versions-mautic-prior-2-16-5-and-3-2-4
+cve: CVE-2020-35124
+branches:
+    '3.2':
+        time:     2021-01-14 17:07:00
```
stof, line 6: these should have a `>= 3.0.0` as well, so that they won't cover the 2.16.5 release.

RCheesley: Both vulnerabilities are affecting both versions, in case that is relevant.

stof: but here, the constraint `<3.2.3` would mark the 2.16.5 release as affected as it matches that constraint. That's not expected to be true, as 2.16.5 is meant to patch them too.

xabbuh: see #524

PR #524, opened 2021-01-18
[mautic/core] fix version constraints

xabbuh:


xabbuh:
ping @RCheesley Can you double-check the change please? :)

xabbuh:
One more thing I was not sure about: Is 3.0.x affected too and should be mark as vulnerable as well or is using any of that versions still fine?

RCheesley:
> One more thing I was not sure about: Is 3.0.x affected too and should be mark as vulnerable as well or is using any of that versions still fine?

Yes indeed, sorry I should have included the 3.0 branch in addition to 3.1 and 3.2. Thanks for the help!

mautic/core/CVE-2020-35124.yaml
```diff
@@ -4,11 +4,11 @@ cve: CVE-2020-35124
 branches:
     '3.2':
         time:     2021-01-14 17:07:00
-        versions: ['<=3.2.3']
+        versions: ['>=3.2.0', '<3.2.4']
     '3.1':
         time:     2021-01-14 17:07:00
```
stof, line 9: this looks weird to me. If the branch is not patched, it should not have a time for the patch

RCheesley: I'm not sure what you mean here - the patch was released at the time on the timestamp - https://github.com/mautic/mautic/releases/tag/3.2.4

stof: but not in the 3.1 branch

RCheesley: Maybe we should only have included the 2.16 and 3.2 branches if the others (3.0 and 3.1) do not get the patch merged back - is this what you mean @stof ?

The 3.1 and 3.0 branches are no longer updated going forward.

stof: either don't include them at all (and put `>=3.0.0` as min version for the 3.2 branch) or put the branch with `time: ~` as being unpatched

RCheesley: OK thanks. The latter part seems to be missing from the contribution documentation (about unpatched branches). Can you make those changes in this PR or am I to make a new PR to address this?

stof (approved):

PR #526, opened 2021-01-19
Add the Enlightn security checker

paras-malhotra:
Just released a [security checker](https://github.com/enlightn/security-checker) package that is similar to the Sensiolabs security checker (now deprecated). I'm adding this to the Readme so that people are aware of this and can use it.

Benefits / differentiation of this package include:
- Can be pulled in with Composer.
- Exposes a PHP API.
- Licensed under the MIT license.

The package is backed with tests and implements HTTP caching while pulling in the Advisories database from Github. 🚀

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,13 +23,18 @@ applications beside manual checks:
 
         local-php-security-checker --path=/path/to/composer.lock
 
- * Use the [Symfony CLI][2] (no PHP dependency, no third-party API calls,
+* Use the [Enlightn Security Checker][2] (can be pulled in using Composer,
+  exposes a PHP API and is licensed under the MIT license):
+
+       php security-checker security:check /path/to/composer.lock
+
+ * Use the [Symfony CLI][3] (no PHP dependency, no third-party API calls,
    checks are done locally on a clone of this repository):
 
         symfony security:check /path/to/composer.lock
 
 **TIP**: If you are using Github, you can use the PHP Security Checker [Github
-Action][3] to automatically check for vulnerabilities when pushing code.
+Action][4] to automatically check for vulnerabilities when pushing code.
 
 Contributing
 ------------
@@ -83,5 +88,6 @@ when you have read-only subtree splits of a main repository), duplicate the
 information in several files.
 
 [1]: https://github.com/fabpot/local-php-security-checker
-[2]: https://symfony.com/download
-[3]: https://github.com/marketplace/actions/the-php-security-checker
+[2]: https://github.com/enlightn/security-checker
+[3]: https://symfony.com/download
+[4]: https://github.com/marketplace/actions/the-php-security-checker


paras-malhotra:
@fabpot, would really appreciate if you could take a few minutes to review this PR. I think it will help many people who are looking for a library with a PHP API or something that can be pulled in via Composer. Thank you.

paras-malhotra:
Hi @fabpot, just wanted to bring this to your notice again. Would really appreciate if you could merge this PR. 

There are many use cases where the current recommended options are not suitable (as highlighted in my first comment). This would provide another implementation option for many folks. Some popular libraries such as PHPQA and GrumPHP are now using the Enlightn security checker for these reasons.

paras-malhotra:
Hi @fabpot, a gentle reminder on this.

The Enlightn Security Checker is now pulled in as a dependency of Drush (and consequently Drupal projects) via https://github.com/drush-ops/drush/pull/4653. We'll soon cross 20k daily downloads with the next Drush RC release. 

I think that's sufficient validation that the Enlightn Security Checker is widely supported by the PHP community and also that the current recommended options don't suffice for many use cases. So, can we please merge this PR?

paras-malhotra:
@naderman is it possible for you to review and merge this PR? Seems like Fabien is busy / not responsive, so was just wondering if some of the other maintainers can look into this?

fabpot:
I'm :-1: to add this reference here. The main reason is that I don't like to add tools as dependencies to projects. That's one of the reason I've coded this tool in Go instead of PHP. I can understand that this is controversial, but I will stick to my gut feeling here.

paras-malhotra:
@fabpot, with a Composer global require, it is not needed to add this as a dependency to project. In fact, that is the recommended option.

paras-malhotra:
To follow up on this, I will detail the Pros and Cons of using the Enlightn Security Checker:

**Cons**: I cannot think of any con really. The installation reason that you mentioned is actually not valid. It doesn't need to be a project dependency. Project dependency installation is an option, not a requirement.

**Pros:**
- **Easy installation via Composer**: People scanning for PHP vulnerabilities are more than likely to have Composer installed and would want an easy installation rather than downloading binaries that are dependent on system architecture (local CLI tool and Symfony CLI).
- **Exposes a PHP API**: Many tools such as Drush, GrumPHP, etc. prefer a PHP library to perform their checks for easy integration with their tools.
- **Licensing**: The Local CLI tool is licensed under AGPL whereas the Enlightn Security Checker is licensed under the MIT license (which is much more permissive).

fabpot:
Thank you for the discussion. Closing now.

paras-malhotra:
@fabpot I don't understand this. Did you go through my comments above? The reason why you didn't want to add this is not applicable to the Enlightn Security Checker. It is not a requirement to pull this in as a project dependency.

@naderman I would request if the other maintainers can look into this. Since Enlightn directly competes with Symfony Insights, I think Fabien might be biased here and would appreciate an independent review of this PR. FriendsOfPHP is meant to be a community organization, not a dictatorship.

fabpot:
WHAT? I think you are going too far now. Biased? Dictatorship? After giving so many years to the community for free? I've even open-source the tool. You have not even understood my objection. I'm very disappointed.

paras-malhotra:
Despite my explanation that your objection was not applicable to the package, you closed this PR without any reason or follow-up comment. You ran Sensiolabs security checker for 8 years with the exact same installation options. Forgive me if I think you're biased but I can't think of any other reason here.

Yes, you gave many years to the community for free but the past doesn't justify the present. Here, it just seems that you're discouraging other open source packages that have perfectly valid use cases and are already widely adopted by the PHP community.

nicolas-grekas:
I personally don't think that tools managed `composer global` are workable in practice. They're too easy to forget to update, etc. I've used `composer global` in the past, with hope, but I think it just doesn't work in practice (no offense to the work done by @naderman and others on the topic!)

I'd much prefer something like https://github.com/composer/composer/issues/9636 (or any of the alternatives discussed there) to become mainstream. As such, I consider @fabpot's still valid.

About "dictatorship" and "bias", this is just not fair. Rejecting this doesn't prevent anyone from using your work @paras-malhotra. It might just not make it as visible as you'd like to, but this is not something that this very repo is aimed at - giving visibility to specific work.

paras-malhotra:
@nicolas-grekas, how is downloading binaries easier to update than Composer global? Also, we have a downloadable PHAR available as well if that's what you prefer. Besides, Sensiolabs security checker had the exact same installation options for 8 years and where were these objections then?

Also, have you gone through the pros listed above? Don't you think they have various advantages over the current recommended options?

I am just asking for an independent review by other maintainers. I am not accusing of @fabpot being biased but I am saying that he **may** have been biased because I don't see any other valid reason to avoid/reject this PR despite me giving an explanation that his objection wasn't valid for this package.

naderman:
@nicolas-grekas You'll find I'm the first one to tell people not to use compoer global at all. When adding that option it was really against my better judgement due to popular demand and there's been nothing but problems with it since.

Ocramius:
Even without composer global installs, it's trivial to `mkdir my-tool && cd my-tool && composer require enlightn/security-checker`, which works pretty much OK.

Given the homepage already lists some tools, I don't see why it is not in line with this package to suggest other libraries/binaries that use it for a good purpose.

Otherwise, I don't see a reason to have things like the security checker and such being there either :shrug:

paras-malhotra:
Here's my take:

1. In my opinion, the installation issue highlighted is not a major one. As @Ocramius suggested, there are several ways around it. It's worth considering that is downloading a binary really any better? A Composer installation may not be perfect but imo it's definitely easier to update than checking the release page and downloading a new binary (specific to system arch).
2. Even if the binary installation method is preferred, we always have the option to download the PHAR.
3. There are several advantages with the Enlightn Security Checker that we should also consider. Firstly, there are many libraries that need PHP APIs to scan PHP vulnerabilities in code. Secondly, an AGPL license is very restrictive and it's not possible to use the Local CLI because of its restrictive licensing policy. In fact, many companies like Google have a strict NO to AGPL and [for good reason](https://opensource.google/docs/using/agpl-policy/).

I don't think it's an either OR situation. There's no taking away that due to licensing and the PHP API, there are several use cases that are simply not possible using the current recommended options. That's why the question is not Local CLI vs Enlightn Security Checker. The question is should we add it as an option to possibly serve other use cases. Can it serve other use cases that currently are not possible with the recommended options?

I would appreciate if there is an independent review from another maintainer who considers all the factors mentioned above.

PR #527, opened 2021-01-21
Add Drupal core SA-CORE-2021-001

eiriksm:


stof:
when using drupal through composer, is Archive_Tar bundled inside `drupal/core` or installed as a composer dependency ? If it is installed by composer, there is no point making an advisory for drupal packages for that. The advisory on the archive_tar library (already added in https://github.com/FriendsOfPHP/security-advisories/blob/master/pear/archive_tar/CVE-2020-36193.yaml) will be reported for the project when using a vulnerable version.

eiriksm:
Hm good point. It's not bundled inside any of those packages, but the recommended way is to use drupal/core-recommended which pins it's dependencies. But as I said, that is another package I guess, which is not covered by this repo

stof:
@eiriksm the dependency pinning is not an issue. you will still get the `pear/archive_tar` package in you composer.lock and so its advisory will be detected when needed.

`drupal/core-recommended` does not appear in this repo because it would not make sense: this is a metapackage so it does not contain any source code. Not having any source code prevents having vulnerable source code (dependencies on vulnerable dependencies are not the point here, as the advisory is set on the dependency)

PR #528, opened 2021-01-22
Update Laravel database advisory

jaylinski:
The previous fix could be bypassed, so new versions were released.

Source:
* https://github.com/laravel/framework/pull/35865#discussion_r561578503
* https://github.com/laravel/framework/pull/35972

### DIFF
--- a/illuminate/database/CVE-2021-21263.yaml
+++ b/illuminate/database/CVE-2021-21263.yaml
@@ -1,14 +1,14 @@
 title:     Unexpected bindings in QueryBuilder
 link:      https://blog.laravel.com/security-laravel-62011-7302-8221-released
-cve:       ~
+cve:       CVE-2021-21263
 branches:
     "6.x":
         time:     2020-01-13 14:35:00
-        versions: ['>=6.0.0', '<6.20.11']
+        versions: ['>=6.0.0', '<6.20.14']
     "7.x":
         time:     2020-01-13 14:37:00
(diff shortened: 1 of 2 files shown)


stof:
I think this should rather be added as separate advisories linking to https://blog.laravel.com/security-laravel-62012-7303-released instead of updating the existing one while keeping the old link.

jaylinski:
@stof I don't know why @taylorotwell didn't update the advisory (https://github.com/advisories/GHSA-3p32-j457-pg5x).

But as you can see in PR https://github.com/laravel/framework/pull/35972, another fix was required to address the security issue. It was tagged with the versions I specified in the this PR.

naderman:
The files were already moved to the correct CVE locations in https://github.com/FriendsOfPHP/security-advisories/pull/529 so this conflicts now - was there any feedback/response on creating an additional advisory / at least updating the announcement blog post?

jaylinski:
@naderman Not from an official side: https://github.com/laravel/framework/pull/35972#issuecomment-767846044

naderman:
@jaylinski Maybe instead of modifying this issue, you can make a separate one for the newer versions which simplify references https://blog.laravel.com/security-laravel-62012-7303-released as a link as that's public confirmation those versions were still vulnerable? That way the existing CVE issue matches what was reported on the issue, but the newer versions are marked insecure via the yet to be created file? Thanks!

PR #532, opened 2021-02-01
Add new Laravel database advisory

jaylinski:
The previous fix (CVE-2021-21263) could be bypassed,
so new versions were released.

Source:
* https://github.com/laravel/framework/pull/35865#discussion_r561578503
* https://github.com/laravel/framework/pull/35972

Closes #528.

### DIFF
--- a/illuminate/database/2021-01-21.yaml
+++ b/illuminate/database/2021-01-21.yaml
@@ -1,0 +1,14 @@
+title:     Unexpected bindings in QueryBuilder
+link:      https://github.com/laravel/framework/pull/35972
+cve:       ~
+branches:
+    "6.x":
+        time:     2020-01-21 15:10:00
+        versions: ['>=6.0.0', '<6.20.14']
+    "7.x":
+        time:     2020-01-21 15:10:00
+        versions: ['>=7.0.0', '<7.30.4']
+    "8.x":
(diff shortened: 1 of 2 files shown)


naderman:
@jaylinski thanks, think this make more sense, however the source/link seems to be some third party blog, rather than laravel itself now?

jaylinski:
@naderman Yeah, that was deliberate, since there is no official blog-post on this security-release. Do you think it is better to include the official advisory? I want to avoid confusion (see https://github.com/FriendsOfPHP/security-advisories/pull/528#issuecomment-765418865).

stof:
maybe use the PR as target of the link instead of using a non-official source (the merged PR is in the official repository)

naderman:
Indeed, that would work for now and we can still update the link if Laravel clarify this in their blog posts further.

jaylinski:
@naderman @stof Good idea. I updated the PR.

stof (approved):

naderman (approved):

naderman:
@jaylinski they added a new advisory now: https://github.com/laravel/framework/security/advisories/GHSA-x7p5-p2c9-phvg can you send another PR for the updated link? Thank you for all the help here!

PR #534, opened 2021-02-03
Use --dir option for symfony security:check

bytehead:
Use `--dir` option for symfony security:check otherwise it will fail:
```bash
 Incorrect usage: Too many arguments
```


bytehead:
Thank you @fabpot!

PR #535, opened 2021-02-03
Simplify regexp

Rudloff:
This does the same thing but with a slighty easier to read syntax:
* `(\\<|\\>)` can be simplified as `([<>])`.
* `{0,1}` can be simplified as `?`.
* Escaping `.` is not required in `[]`.
* Escaping `-`  is not required outside `[]`.

### DIFF
--- a/validator.php
+++ b/validator.php
@@ -63,7 +63,12 @@ final class Validate extends Command
         };
 
         $isAcceptableVersionConstraint = function ($versionString) {
-            return (bool) preg_match('/^[<>]=?(([1-9]\d*)|0)(\.(([1-9]\d*)|0))*(-(alpha|beta|rc|p(atch)?)[1-9]\d*)?$/', $versionString);
+            return (bool) preg_match('/^
+                [<>]=?
+                (([1-9]\d*)|0)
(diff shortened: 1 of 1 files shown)


rgson:
I think it can be further simplified.
- `([<>])(=)?` could be `[<>]=?` because the captured matches don't seem to be used anyway.
- `[\d.\-]` could be `[\d.-]` because escaping `-` isn't required when it's the last character in a character class.

garas:
Long patterns can be split to multiple lines with `x` flag to separate parts to be easier to read:
```
/^
    [<>]=? 
    (([1-9]\d*)|0)
    (\.(([1-9]\d*)|0))*
    (-(alpha|beta|rc|p(atch)?)[1-9]\d*)?
$/x
```

stof:
> * `[\d.\-]` could be `[\d.-]` because escaping `-` isn't required when it's the last character in a character class.

From a maintenance PoV, I would keep the escaping here. It avoids breaking it by mistake if we later add more things in the character class (and we add it at the end)

stof (approved):

PR #536, opened 2021-02-15
Adds Record For facade/ignition RCE: CVE-2021-3129

freshleafmedia:
This PR adds a record for the facade/ignition RCE vulnerability.

- https://www.ambionics.io/blog/laravel-debug-rce
- https://github.com/facade/ignition/pull/334
- https://github.com/facade/ignition/pull/353

### DIFF
--- a/facade/ignition/CVE-2021-3129.yaml
+++ b/facade/ignition/CVE-2021-3129.yaml
@@ -1,0 +1,8 @@
+title:     Remote code execution
+link:      https://github.com/facade/ignition/pull/334
+cve:       CVE-2021-3129
+branches:
+    "master":
+        time:     2020-11-17 09:18:00
+        versions: ['<=2.5.1']
+reference: composer://facade/ignition
--- a/facade/ignition/CVE-2021-3129.yaml
+++ b/facade/ignition/CVE-2021-3129.yaml
@@ -4,5 +4,8 @@ cve:       CVE-2021-3129
 branches:
     "master":
         time:     2020-11-17 09:18:00
-        versions: ['<=2.5.1']
+        versions: ['<=2.5.1', '>=2.0']
+    "1.x":
+        time:     2021-02-13 10:47:03
+        versions: ['<=1.16.13']
 reference: composer://facade/ignition
--- a/facade/ignition/CVE-2021-3129.yaml
+++ b/facade/ignition/CVE-2021-3129.yaml
@@ -5,7 +5,4 @@ branches:
     "master":
         time:     2020-11-17 09:18:00
         versions: ['<=2.5.1']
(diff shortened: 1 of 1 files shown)


naderman:
Am I missing something? If I look at https://github.com/facade/ignition/tree/1.16.14 it doesn't seem to actually contain the fix? Did something go wrong with tagging there @freekmurze ?

freshleafmedia:
@naderman Is this not the fixed code: https://github.com/facade/ignition/blob/1.16.14/src/Solutions/MakeViewVariableOptionalSolution.php#L74-L90

naderman:
@freshleafmedia It is, but it looks to me like it's actually a version of 2.5 containing the fix that was accidentally taged as 1.16? The rest of the code is 2.5 now in the 1.16 tag too? https://github.com/facade/ignition/blob/1.16.14/CHANGELOG.md

naderman:
Regarding your invalid bound change, it was fine to add v1 with the constraint, you just have to then add a lower bound for master at >=2.0-dev because otherwise the two branches overlap?

freshleafmedia:
Really? I was getting an `Invalid branch name "v1".` error

naderman:
That's a poorly worded error message I think. It's not the name of the branch that's invalid, it's an invalid branch definition with the name "v1":

Screenshot_20210215_111809

freshleafmedia:
I'm afraid not, It's two separate error messages. If I add a lower version constraint it shows the error:

```
                                                                                                                        
 [ERROR] Found 1 issue in 1 file.                                                                                       
                                                                                                                        

+------------------------------------+---------------------------+
| File                               | Issues                    |
+------------------------------------+---------------------------+
| facade/ignition/CVE-2021-3129.yaml | Invalid branch name "v1". |
+------------------------------------+---------------------------+

```

freshleafmedia:
Looking at the validator it is using the regex `^([\d\.\-]+(\.x)?(\-dev)?|master)$` so only a version number, a version number suffixed by `-dev` or just `master` on its own are valid options

naderman:
@freshleafmedia ah sorry, the name is indeed invalid, I suppose "1.x" would be an appropriate branch name to store the info on the 1.x version range, maybe @stof can shed some light on this?

stof:
I think `1.x` is the way to go indeed (assuming that there is indeed a valid fixed version for 1.x)

valentijnscholten:
How far is this from getting merged? There seems to be active probing for this vulnerability according to greynoise: 
image
https://twitter.com/nathanqthai/status/1367130234663940110

naderman:
I see that the 1.16.14 release has been fixed in the meantime, so merging it now.

PR #544, opened 2021-03-22
Add CVE-2021-27908

alanhartless:


mautic/core/CVE-2021-27908.yaml
```diff
@@ -0,0 +1,20 @@
...
+        versions: ['>=3.2.0', '<3.3.0']
+    '3.1':
+        time: ~
+        versions: ['>=3.1.0', '<3.2.0']
+    '3.0':
+        time: ~
+        versions: ['>=3.0.0', '<3.1.0']
+    '2.16':
+        time: ~
+        versions: ['>=2.0.0', '<=2.16.5']
```
xabbuh, line 19: Is the upper boundary right? I don't see anything in the related advisory that mentions the 2.16 release.

naderman: The advisory actually states anything `<3.3.2` without excluding 2.x so this seems like a reasonable assumption, although at that point I wonder why 1.x is excluded here?

alanhartless: Right, the advisory is all versions prior to 3.3.2 (patched only in 3.3.2). 2.x is no longer supported with the latest and last release being 2.16.5. Should we do something different here? 

1.x is so old that we didn't even think to include them since starting to participate here. Nor do we have 1.x branches. If needed, I supposed we could create a branch from the latest 1.x tag and add it to the branches list for the advisories? Let me know. 

Thanks!

xabbuh: I think if 2.x is no longer supported, I would set the upper bound to `<3.0`.

stof: or maybe just remove the `>=3.0.0` condition in the 3.0 branch, so that the range also covers all 2.x and 1.x versions.

alanhartless: Which direction should I take? 

```
branches:
    '3.3':
        time: 2021-03-22 20:15:00
        versions: [ '>=3.3.0', '<3.3.2' ]
    '3.2':
        time: ~
        versions: ['>=3.2.0', '<3.3.0']
    '3.1':
        time: ~
        versions: ['>=3.1.0', '<3.2.0']
    '3.0':
        time: ~
        versions: ['>=3.0.0', '<3.1.0']
    '2.16':
        time: ~
        versions: [ '>=2.0.0', '<3.0.0' ]
```

or

```
branches:
    '3.3':
        time: 2021-03-22 20:15:00
        versions: [ '>=3.3.0', '<3.3.2' ]
    '3.2':
        time: ~
        versions: ['>=3.2.0', '<3.3.0']
    '3.1':
        time: ~
        versions: ['>=3.1.0', '<3.2.0']
    '3.0':
        time: ~
        versions: ['<3.1.0']
```

We still have a 2.16 branch for archival purposes if that influences things.

alanhartless: @xabbuh @stof thoughts on the above?

xabbuh: I'd go with your second example.

stof:
@alanhartless any chance to update your PR ?

mautic/core/CVE-2021-27908.yaml
alanhartless: @xabbuh updated. Thanks!

alanhartless:
@stof yes; I apologize for the delay. It's been a busy couple weeks. It's been updated.

stof (approved):

PR #554, opened 2021-04-15
Add neos/* advisories

kdambekalns:
This adds advisories for neos/* under the neos vendor name, those
packages replace the ones from typo3/*.


stof (approved):
I'll trust you that the list is correct. I haven't checked them in detail one by one.

PR #572, opened 2021-06-16
Add phpmailer/phpmailer CVE-2021-3603 and CVE-2021-34551

Synchro:


phpmailer/phpmailer/CVE-2021-34551.yaml
```diff
@@ -0,0 +1,8 @@
+title:     RCE affecting Windows hosts via UNC paths to translation files
+link:      https://github.com/PHPMailer/PHPMailer/releases/tag/v6.5.0
+cve:       CVE-2021-34551
+branches:
+    6.x:
+        time:     2021-06-16 16:20:00
+        versions: ['<=6.4.1']
```
stof, line 7: I'd rather use `<6.5.0` as 6.5.0 is the patched version (it is equivalent *if there is no other 6.4.x version out there*, but my suggestion avoids that need by relying on the patched version rather than on the last unpatched version)

Synchro: Done

stof (approved):

PR #578, opened 2021-07-29
Add advisory for CVE-2021-32697 in neos/form

kdambekalns:


xabbuh:
As far as I understand the related advisory every release from 1.2.0 to 4.3.2 is affected. Shouldn't this be reflected here?

kdambekalns:
Well, those versions are no longer supported, but I amended the affected versions. Hope that does the job!

PR #586, opened 2021-09-03
Add CVE-2021-23420 in codeception/codeception

particleflux:
See:
* https://github.com/advisories/GHSA-4574-qv3w-fcmg
* https://nvd.nist.gov/vuln/detail/CVE-2021-23420

Again, this is untested, as I still haven't found a way to test these with a local repository

### DIFF
--- a/codeception/codeception/CVE-2021-23420.yaml
+++ b/codeception/codeception/CVE-2021-23420.yaml
@@ -1,0 +1,14 @@
+title:     Deserialization of Untrusted Data
+link:      https://github.com/advisories/GHSA-4574-qv3w-fcmg
+cve:       CVE-2021-23420
+branches:
+    "4.1":
+        time:     2021-08-06 17:20:00
+        versions: ['>=4.1.0', '<4.1.22']
+    "4.0":
+        time:     ~
+        versions: ['>=4.0.0', '<4.1.0']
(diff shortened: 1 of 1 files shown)


particleflux:
Please hold, gonna push an update to fix 4.0 detection

particleflux:
Okay, seems to work against 3.x and 4.x now :)

codeception/codeception/CVE-2021-23420.yaml
```diff
@@ -0,0 +1,14 @@
+title:     Deserialization of Untrusted Data
+link:      https://github.com/advisories/GHSA-4574-qv3w-fcmg
+cve:       CVE-2021-23420
+branches:
+    "4.1":
+        time:     2021-08-06 17:20:00
+        versions: ['>=4.1.0', '<4.1.22']
+    "4.0":
+        time:     2021-08-06 17:20:00
```
stof, line 9: this should be `~` (i.e. `null`) as it is not patched in 4.0

stof (changes requested):

codeception/codeception/CVE-2021-23420.yaml
particleflux: Alright, updated

PR #591, opened 2021-10-05
Add security advisories for TYPO3 v11.5 LTS release

ohader:


typo3/cms-core/CVE-2021-41113.yaml
```diff
@@ -0,0 +1,8 @@
+title: 'TYPO3-CORE-SA-2021-014: Cross-Site-Request-Forgery in Backend URI Handling'
+link: 'https://typo3.org/security/advisory/typo3-core-sa-2021-014'
+cve: CVE-2021-41113
+branches:
+    11.x:
+        time: '2021-10-05 11:02:10'
+        versions: ['>=11.0.0', '<11.5.0']
```
xabbuh, line 7: `>=11.2.0` (as in the linked blog post)?

ohader: Good catch! 👍 
These files are automatically generated and it seems we still have a bug there.

PR #594, opened 2021-10-20
fix: add CVE number for illuminate/database guard bypass vulnerability

G-Rath:
Didn't know if you'd like the file renamed too so have left it as-is for now (in part since I could use the GH editor 😅).

CVE-2020-24940
https://nvd.nist.gov/vuln/detail/CVE-2020-24940
https://www.cvedetails.com/cve/CVE-2020-24940/


naderman:
Looks like the test now fails due to filename cve id mismatch, so I suppose renaming the file is the way to go ;-)

PR #595, opened 2021-10-20
fix: add CVE number for league/flysystem advisory

G-Rath:
CVE-2021-32708


fabpot:
File names should not be relied upon, so it's safe to rename files.

PR #596, opened 2021-11-12
added: dapphp/securimage XSS vulnerabilities

8ctopus:
My first pull request to the repo, please check if correct.


naderman:
From what I understand that XSS vulnerabilities only existed in examples, so they do not really apply to users of the library?

8ctopus:
@naderman Thank you for your comment. I went to check and you're right, it concerned only the samples.

PR #599, opened 2021-11-17
Export advisories in OSV format

### LINKED ISSUE #576: [Discussion] Adopt OSV unified vulnerability schema for open source
### Introduction

Google recently published it's OSV unified vulnerability schema for open source: https://security.googleblog.com/2021/06/announcing-unified-vulnerability-schema.html

> OSV is a vulnerability database and triage infrastructure for open source projects aimed at helping both open source maintainers and consumers of open source.

> For open source maintainers, OSV's automation helps reduce the burden of triage. Each vulnerability undergoes automated bisection and impact analysis to determine precise affected commit and version ranges.

> For open source consumers, OSV provides an API that lets users of these projects query whether or not their versions are impacted.

### Discussion

It like the idea of having a unified schema for open source vulnerabilities, so I think it would be nice if this repository could adopt the new OSV schema.

I see the following benefits:

* No more discussion about the format (since it now follows a standard)   
  Relates to #537, #496, #465
* The unified schema will make it easy for other libraries to use this vulnerability source as a data-feed, since it follows a schema also used for other languages   
  Example: https://github.com/pypa/advisory-db/blob/main/vulns/aiohttp/PYSEC-2021-76.yaml
* This vulnerability feed can be made accessible via https://osv.dev/ API

I'm looking forward for your input. :octocat: 

The format-change should be pretty straight forward. I'll open a PR if this proposal receives positive feedback.

jaylinski:
Fixes #576

### Possible improvements

* [ ] Validate generated JSON against spec (https://github.com/ossf/osv-schema/blob/main/validation/schema.json)
* [ ] Validate uniqueness of IDs


export-osv.php
```diff
@@ -0,0 +1,119 @@
...
+function getEvents(array $branches): array
+{
+    $events = [];
+
+    foreach (array_column($branches, 'versions') as $branch) {
+        if (count($branch) === 2) {
+            array_push($events, ['introduced' => $branch[0]]);
+            array_push($events, ['fixed' => $branch[1]]);
+        } else {
+            array_push($events, ['fixed' => $branch[0]]);
```
jaylinski, line 70: :warning: This is not correct, because this DB only specifies ranges that are affected, but not the fixed version.

Maybe we can just convert versions like `<8.22.1` to `8.22.1`j

naderman: How useful is OSV if we cannot list fixed versions at all?

export-osv.php
```diff
@@ -0,0 +1,119 @@
...
+}
+
+function getEvents(array $branches): array
+{
+    $events = [];
+
+    foreach (array_column($branches, 'versions') as $branch) {
+        if (count($branch) === 2) {
+            array_push($events, ['introduced' => $branch[0]]);
+            array_push($events, ['fixed' => $branch[1]]);
```
stof, line 68: be careful. We have cases like `>= 1.3.0, <2.0` where things are not fixed in 2.0 (if another branch has `>=2.0` in its affected constraints)

export-osv.php
```diff
@@ -0,0 +1,119 @@
...
+    ];
+}
+
+function getEvents(array $branches): array
+{
+    $events = [];
+
+    foreach (array_column($branches, 'versions') as $branch) {
+        if (count($branch) === 2) {
+            array_push($events, ['introduced' => $branch[0]]);
```
stof, line 67: `$branch[0]` is not a version but a constraint

export-osv.php
```diff
@@ -0,0 +1,119 @@
...
+            ],
+        ],
+        'references' => [
+            array_key_exists('link', $advisory) ? [
+                'type' => 'ADVISORY',
+                'url' => $advisory['link'],
+            ] : null,
+            [
+                'type' => 'PACKAGE',
+                'url' => 'https://packagist.org/packages/' . $package,
```
stof, line 55: this looks broken for packages using a custom repository

export-osv.php
```diff
@@ -0,0 +1,119 @@
...
+            'ranges' => [
+                'type' => 'SEMVER',
+                'events' => getEvents($advisory['branches']),
+            ],
+        ],
+        'references' => [
+            array_key_exists('link', $advisory) ? [
+                'type' => 'ADVISORY',
+                'url' => $advisory['link'],
+            ] : null,
```
stof, line 52: putting `null` in the array looks wrong to me

export-osv.php
jaylinski: @naderman I worked around this by retrieving all package tags from the Packagist API and comparing the ranges with the tags by calling `Semver::satisfies($version, implode(' ', $constraints))`.

Example: https://github.com/jaylinski/security-advisories/blob/osv/packagist/CVE-2021-43608.json

What do you think about it?

export-osv.php
jaylinski: That's true. I reworked it, see https://github.com/FriendsOfPHP/security-advisories/pull/599#discussion_r752573875.

export-osv.php
jaylinski: That's true. I reworked it, see https://github.com/FriendsOfPHP/security-advisories/pull/599#discussion_r752573875.

export-osv.php
jaylinski: I'm now skipping packages that have a custom or no composer-repository.

export-osv.php
jaylinski: True. Since the validator enforces links anyway, I just skipped that check.

export-osv.php
naderman: Would it make more sense to add this to packagist.org's security advisories API as a new output format (which pulls data from this repo here)? As you'd have access to live package data there?

jaylinski: Yes, definitely. I wasn't ware of the packagist.org security advisories API.

Should I open a PR at the packagist.org repo?

naderman: Sure, go ahead! :-)

It's pretty simple right now, just a way to list advisories for a set of packages, see https://packagist.org/apidoc#list-security-advisories

icanhazstring:
Any new update here to get the fixed version into the list as well?

jaylinski:
@icanhazstring The PHP and OSV vulnerability schemes don't have a `fixed`-field. Only the `affected` versions are listed.

(Or maybe I'm misunderstanding your question?)

icanhazstring:
@jaylinski was referring to the comment from @naderman about the Packagist advisory api about the fixed version.

Or is it somewhat possible to get security issue listed with affected version and the next which fixes it?

ohader:
Just wanted to leave that here (probably you knew it already):
https://github.com/github/advisory-database/blob/b07a1c25e2ec4fe59bf3dae2c6b7db3b02f4ae75/advisories/github-reviewed/2022/04/GHSA-x7cr-6qr6-2hh6/GHSA-x7cr-6qr6-2hh6.json
* gives an example of `fixed` and `introduced` nested in `affected`
* OSV items do exist already in GitHub's advisory DB - can we some "synchronize" multiple sources (did not think about the details & consequences, yet)

naderman:
@ohader The packagist.org API already synchronizes and merges github's db and friendsofphp, e.g. see here: https://packagist.org/packages/guzzlehttp/guzzle/advisories?version=6278149

It would really just need someone to build an OSV compatible output for the data we collect there to have an OSV database for PHP.

wojtekxtx:
@naderman I woul;d not use `packagist.org` code. 

@all: OSV may be the solution, but knowing Google atitude towards users' data, Id look elsewhere.

Raen111 (approved):

PR #601, opened 2021-12-08
Add CVE-2021-43808 laravel/framework

OpenPrunus:


stof:
the advisory should also be added for the `illuminate/view` package, for projects using the subtree packages rather than the monorepo package.

OpenPrunus:
I add also on the illuminate/view package

illuminate/view/CVE-2021-43808.yaml
```diff
@@ -0,0 +1,14 @@
...
+  "6.x":
+    time:     ~
+    versions: ['>=6.0.0', '<6.20.42']
+  "7.x":
+    time:     ~
+    versions: ['>=7.0.0', '<7.30.6']
+  "8.x":
+    time:     ~
+    versions: ['>=8.0.0', '<8.75.0']
+reference: composer://laravel/framework
```
stof, line 14: should be `illuminate/view` here

OpenPrunus: Oops sorry. Thank you

stof (approved):

illuminate/view/CVE-2021-43808.yaml
```diff
@@ -0,0 +1,14 @@
+title:     Possible cross-site scripting (XSS) vulnerability in the Blade templating engine
+link:      https://github.com/laravel/framework/security/advisories/GHSA-66hf-2p6w-jqfw
+cve:       CVE-2021-43808
+branches:
+  "6.x":
+    time:     ~
+    versions: ['>=6.0.0', '<6.20.42']
```
xabbuh, line 7: What about versions below 6.0? Are they not affected?

stof: The github advisory marks them as affected.

stof: this means that the lower bound should be removed here

PR #604, opened 2022-01-21
Add CVE-2022-0323 mustache/mustache

bobthecow:
https://github.com/bobthecow/mustache.php/releases/tag/v2.14.1

### DIFF
--- a/mustache/mustache/CVE-2022-0323.yaml
+++ b/mustache/mustache/CVE-2022-0323.yaml
@@ -1,0 +1,8 @@
+title:     Possible RCE when rendering untrusted user templates
+link:      https://github.com/bobthecow/mustache.php/releases/tag/v2.14.1
+cve:       CVE-2022-0323
+branches:
+    2.x:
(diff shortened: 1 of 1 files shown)


mustache/mustache/CVE-2022-0323.yaml
```diff
@@ -0,0 +1,8 @@
+title:     Possible RCE when rendering untrusted user templates
+link:      https://github.com/bobthecow/mustache.php/releases/tag/v2.14.1
+cve:       CVE-2022-0323
+branches:
+    2.x:
+        time:     2022-01-21 06:21:37
+        versions: ['>=2.0.0', '<2.14.1']
```
stof, line 7: was 1.x safe from that ? If no, I would remove the lower bound

bobthecow: v1 was unaffected. IIRC it also predates Composer :)

stof (approved):

PR #605, opened 2022-01-23
Add CVE-2017-15367

dfranco:


bacula-web/bacula-web/CVE-2017-15367.yaml
```diff
@@ -0,0 +1,8 @@
+title: "Bacula-web before 8.0.0-rc2 is affected by multiple SQL Injection vulnerabilities"
+link: https://github.com/bacula-web/bacula-web/releases/tag/v8.0.0-rc.2
+cve: CVE-2017-15367
+branches:
+    master:
+        time:     2018-03-02 16:41:32
+        versions: ['>8.0.0', '<8.0.0']
```
stof, line 7: this won't match any version when applying the range

stof:
Side note: as bacula-web is not installed as a package in projects (the doc says to use that package as a template with `composer create-project` or by downloading the archive), the tooling won't be able to warn based on that advisory anyway.

bacula-web/bacula-web/CVE-2017-15367.yaml
dfranco: Thanks for your feedback.

I tried to define something like ['>8.0.0-rc.1', '<8.0.0-rc.2'] or ['<8.0.0-rc.2'].
But trying to validate my commit I received with the error message
`Version constraint "is not in an acceptable format"`

dfranco:
> Side note: as bacula-web is not installed as a package in projects (the doc says to use that package as a template with `composer create-project` or by downloading the archive), the tooling won't be able to warn based on that advisory anyway.

Good point. Do you mean even using `composer create-project` won't warn users if they're using a version with vulnerabilities ?

stof:
@dfranco composer itself does not show any warning about security advisories for now (there is https://github.com/composer/composer/issues/10329 about it).
and btw, in your case of using `create-project` as the installation tool by making users create a project based on your code, a warning in `create-project` will only help if your latest version is vulnerable at the time users install it (in which case, there is nothing they can do anyway, as it means there is no patched version). Vulnerabilities are generally discovered and registered as advisories later (and so users would have to be warned that the version of bacula-web from which they created a project 3 months ago is now known to be vulnerable)

dfranco:
> @dfranco composer itself does not show any warning about security advisories for now (there is [composer/composer#10329](https://github.com/composer/composer/issues/10329) about it). and btw, in your case of using `create-project` as the installation tool by making users create a project based on your code, a warning in `create-project` will only help if your latest version is vulnerable at the time users install it (in which case, there is nothing they can do anyway, as it means there is no patched version). Vulnerabilities are generally discovered and registered as advisories later (and so users would have to be warned that the version of bacula-web from which they created a project 3 months ago is now known to be vulnerable)

Understood, thanks for your valuable feedback

stof:
Given that this advisory in this DB won't actually allow notifying the bacula-web affected users, and that this repo currently does not support advisories affecting only pre-release versions, maybe the PR should be closed. What do you think @fabpot ?

naderman:
Closing this as per @stof's suggestion as it doesn't seem useful.

dfranco:
Thanks for your feedbacks @stof & @naderman .

Even if it doesn't seem's useless, at least some versions of my project would be shown as "deprecated" from Packagist website.

Let me know if there's something I can do on my side.

Thanks again

PR #616, opened 2022-03-14
Add CVE-2021-43676

Rudloff:
See https://github.com/advisories/GHSA-mh9j-v6mq-pfch.

There is no fix yet so we might have to update the YAML later.

### DIFF
--- a/matyhtf/framework/CVE-2021-43676.yaml
+++ b/matyhtf/framework/CVE-2021-43676.yaml
@@ -1,0 +1,8 @@
+title: Path manipulation
+link: https://github.com/advisories/GHSA-mh9j-v6mq-pfch
+cve: CVE-2021-43676
+reference: composer://matyhtf/framework
+branches:
+    master:
+        time: null
+        versions: [ '<=3.0.5' ]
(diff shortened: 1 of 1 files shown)


matyhtf/framework/CVE-2021-43676.yaml
```diff
@@ -0,0 +1,8 @@
+title: Path manipulation
+link: https://github.com/advisories/GHSA-mh9j-v6mq-pfch
+cve: CVE-2021-43676
+reference: composer://matyhtf/framework
+branches:
+    '3.0':
+        time: null
```
stof, line 7: this should have the date set to the time at which the fix was merged

matyhtf/framework/CVE-2021-43676.yaml
```diff
@@ -0,0 +1,8 @@
...
+        versions: [ '<=3.0.5' ]
```
stof, line 8: we generally use `<3.0.6`, mentioning the version that provides the patch rather than the previous version.

stof (changes requested):

matyhtf/framework/CVE-2021-43676.yaml
Rudloff: As I said, there is not fix yet.

stof: if there were no fix yet, `<=3.0.5` as range would have been wrong too

PR #622, opened 2022-03-20
CVE-2022-24775: Guzzle PSR-7

Nyholm:
I just requested a CVE from Github. I guess they will come back next week.

### DIFF
--- a/guzzlehttp/psr7/GHSA-q7rv-6hp3-vh96.yaml
+++ b/guzzlehttp/psr7/GHSA-q7rv-6hp3-vh96.yaml
@@ -1,0 +1,12 @@
+title:     Inproper parsing of HTTP headers
+link:      https://github.com/guzzle/psr7/releases/tag/2.1.1
+cve:       GHSA-q7rv-6hp3-vh96
+branches:
+    master:
+        time:     2022-03-20 13:44:00
+        versions: ['>=2', '<2.1.1']
+    1.x:
+        time:     2022-03-20 13:45:00
+        versions: ['<1.8.4']
+
+reference: composer://guzzlehttp/psr7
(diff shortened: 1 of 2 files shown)


GrahamCampbell (approved):

guzzlehttp/psr7/GHSA-q7rv-6hp3-vh96.yaml
```diff
@@ -0,0 +1,12 @@
+title:     Inproper parsing of HTTP headers
+link:      https://github.com/guzzle/psr7/releases/tag/2.1.1
```
naderman, line 2: Can you link to the advisory instead as it has more info?

```suggestion
link:      https://github.com/guzzle/psr7/security/advisories/GHSA-q7rv-6hp3-vh96
```

naderman (changes requested):

guzzlehttp/psr7/GHSA-q7rv-6hp3-vh96.yaml
```diff
@@ -0,0 +1,12 @@
...
+cve:       GHSA-q7rv-6hp3-vh96
```
naderman, line 3: This is not a CVE id. Please do not list one if there isn't one. I think @nicolas-grekas previously used one with XXX placeholders which i can live with, but this is going to cause issues with scripts processing this automatically. Should probably add something to the CI script for this? cc @stof

naderman:
Please use an appropriate filename as per https://github.com/FriendsOfPHP/security-advisories#contributing if there is no cve I'd yet, use the date format and you can rename the file once there is a cve id.

guzzlehttp/psr7/GHSA-q7rv-6hp3-vh96.yaml
stof: @naderman see #623

Nyholm:
Thank you for the quick review. I got a CVE now and the PR is updated

guzzlehttp/psr7/GHSA-q7rv-6hp3-vh96.yaml
Nyholm: Thank you. I will remember this for next time

stof (approved):

Nyholm:
Thank you for merging b

PR #628, opened 2022-05-06
CVE-2022-24899

Toflar:


Toflar:
That's the monorepo, we haven't done that for any previous advisories, see https://github.com/FriendsOfPHP/security-advisories/tree/master/contao. Shall I add them all in another PR?

stof:
@Toflar as your monorepo is published on Packagist (and so can be used directly instead of using the subtree splits), it is needed (see `symfony/symfony` having all advisories too).
Older advisories should indeed also be applied to it.

Toflar:
Based https://github.com/FriendsOfPHP/security-advisories/pull/629 on top of this one. Just basically copied all the advisories and replaced the `composer://` path. I think that's it?

stof:
> Just basically copied all the advisories and replaced the `composer://` path. I think that's it?

yeah, that's all what's needed as versions are in sync.

Toflar:
Thanks for the support @stof ❤️

PR #635, opened 2022-06-14
Add brotkrueml/schema

brotkrueml:


brotkrueml/schema/2022-06-14-1.yaml
```diff
@@ -0,0 +1,11 @@
+title: Cross-Site Scripting
+link: https://typo3.org/security/advisory/typo3-ext-sa-2022-012
+cve: ~
```
naderman, line 3: according to the link above CVE id is `CVE-2022-33154` and this should also be used for the filename then.

naderman (changes requested):

brotkrueml/schema/2022-06-14-1.yaml
ohader: @brotkrueml Can you rename the file `2022-06-14-1.yaml` to `CVE-2022-33154.yaml` please? Thx in advance!

ohader (approved):

PR #636, opened 2022-06-14
Add security advisories for TYPO3's June 2022 releases

ohader:
https://typo3.org/article/typo3-11511-and-10429-security-releases-published

### DIFF
--- a/typo3/cms-core/CVE-2022-31049.yaml
+++ b/typo3/cms-core/CVE-2022-31049.yaml
@@ -1,0 +1,11 @@
+title: 'TYPO3-CORE-SA-2022-004: Cross-Site Scripting in Frontend Login Mailer'
+link: 'https://typo3.org/security/advisory/typo3-core-sa-2022-004'
+cve: CVE-2022-31049
+branches:
+    10.x:
+        time: '2022-06-14 07:11:45'
+        versions: ['>=10.0.0', '<10.4.29']
+    11.x:
+        time: '2022-06-14 07:12:40'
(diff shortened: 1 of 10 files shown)


typo3/cms-core/CVE-2022-31049.yaml
```diff
@@ -0,0 +1,11 @@
+title: 'TYPO3-CORE-SA-2022-004: Cross-Site Scripting in Frontend Login Mailer'
+link: 'https://typo3.org/security/advisory/typo3-core-sa-2022-004'
+cve: CVE-2022-31049
+branches:
+    10.x:
+        time: '2022-06-14 07:11:45'
+        versions: ['>=10.0.0', '<10.4.29']
```
stof, line 7: what about the 9.5.35 version that is also mentioned in the advisory ?

ohader: Version 9.5.35 is non-public - it's part of the ELTS (extended LTS) program and is a paid service. It's part of the advisories to notify users that also those ELTS version have been updated. However, we as community security team don't really have an insight into those releases.

References
* https://typo3.com/services/extended-support-elts (paid ELTS subscription)
* https://get.typo3.org/list/version/9 (all v9 releases - see when community switched to `ELTS` state end of 2021)

stof: ah, I see that this version is not published to Packagist either. So fine with me.

stof (approved):

PR #640, opened 2022-07-05
CVE-2022-29221

asiminiceanu:
https://github.com/smarty-php/smarty/security/advisories/GHSA-634x-pc3q-cf4c


stof:
Note that the Smarty github advisory misses the `>= 4.0` part in the affected version. So it would mark 3.1.45 as affected due to the `<4.1.1` constraint.

stof (approved):

asiminiceanu:
@stof more info https://nvd.nist.gov/vuln/detail/CVE-2022-29221

stof:
The CVE and the advisory submitted here both contain the proper `from 4.0 to 4.1.1`. But the github advisory does not.

PR #643, opened 2022-07-25
Drupal core - Critical - Arbitrary PHP code execution - SA-CORE-2022-014

barryvdh:
https://www.drupal.org/sa-core-2022-014

I see that a lot of security advisories were missing, but I just added the most recent. Also marked the last 8.9 branch + 9.0/9.1/9.2 as insecure because they are no longer covered in the security process. Combined with https://github.com/FriendsOfPHP/security-advisories/blob/master/drupal/core/CVE-2021-33829.yaml that should only keep 9.3.19 and 9.4.3 als secure options.


barryvdh:
Added a second CVE (from the same date), which also includes the 7.x branch, so that is now also covered.

stof (approved):

PR #644, opened 2022-08-02
Add advisory for CVE-2022-31109 in laminas/laminas-diactoros.

beloglazov91:
Add advisory for CVE-2022-31109 in laminas/laminas-diactoros.


Krilo89:
Would be really awesome to merge this PR, because right now our Drupal projects don't get updates.

naderman:
Does the issue really also impact the 1.x version range?

naderman:
Looks like it, so merged.

PR #646, opened 2022-08-08
Add CVE-2022-35943 codeigniter4/shield

kenjis:
I got the error:
> Version constraint "<1.0.0-beta.2" is not in an acceptable format.

Don't you accept advisories for pre-releases?

### DIFF
--- a/codeigniter4/shield/CVE-2022-35943.yaml
+++ b/codeigniter4/shield/CVE-2022-35943.yaml
@@ -1,0 +1,8 @@
+title:     "CVE-2022-35943: SameSite Attackers may Bypass the CSRF Protection"
+link:      https://github.com/codeigniter4/shield/security/advisories/GHSA-5hm8-vh6r-2cjq
+cve:       CVE-2022-35943
+branches:
+    master:
+        time:     2022-08-07 15:20:07
+        versions: ['<1.0.0-beta.2']
+reference: composer://codeigniter4/shield


MGatner:
The regex in **validator.php** does not allow for a `.` between "beta" and "2":
```php
return (bool) preg_match('/^
                [<>]=?
                (([1-9]\d*)|0)
                (\.(([1-9]\d*)|0))*
                (-(alpha|beta|rc|p(atch)?)[1-9]\d*)?
            $/x', $versionString);
```
... despite this being a valid Semver:

> A pre-release version MAY be denoted by appending a hyphen and a series of dot separated identifiers immediately following the patch version. Identifiers MUST comprise only ASCII alphanumerics and hyphens [0-9A-Za-z-]. Identifiers MUST NOT be empty. Numeric identifiers MUST NOT include leading zeroes. Pre-release versions have a lower precedence than the associated normal version. A pre-release version indicates that the version is unstable and might not satisfy the intended compatibility requirements as denoted by its associated normal version. Examples: 1.0.0-alpha, 1.0.0-alpha.1, 1.0.0-0.3.7, 1.0.0-x.7.z.92, 1.0.0-x-y-z.–.

Notice in the "Examples": `1.0.0-alpha.1`

Ocramius:
TBH, advisories for non-stable releases are not that relevant, since pre-release versions shouldn't be used in a production environment anyway.

PR #647, opened 2022-08-17
Add CVE-2022-30429.yaml for neos/neos

kdambekalns:


neos/neos/CVE-2022-30429.yaml
```diff
@@ -0,0 +1,38 @@
+title:     XSS in various backend modules
+link:      https://www.neos.io/blog/xss-in-various-backend-modules.html
+cve:       CVE-2022-30429
+branches:
+    3.3.x:
+        time:     2022-05-18 17:00:00
```
stof, line 6: the time should be `null` (represented as `~` in YAML) as the branch has not been patched (same for other unpatched branches)

stof (changes requested):

neos/neos/CVE-2022-30429.yaml
kdambekalns: Ah, thanks! Didn't know that…

stof (approved):

PR #648, opened 2022-09-13
Add security advisories for TYPO3's September 2022 releases

ohader:


typo3/cms/CVE-2022-36108.yaml
```diff
@@ -0,0 +1,11 @@
+title: 'TYPO3-CORE-SA-2022-010: Cross-Site Scripting in &lt;f:asset.css&gt; view helper'
```
ohader, line 1: I'm not sure about the encoding here? Should it be plain instead?

typo3/cms-core/CVE-2022-36108.yaml
```diff
@@ -0,0 +1,11 @@
+title: 'TYPO3-CORE-SA-2022-010: Cross-Site Scripting in &lt;f:asset.css&gt; view helper'
```
ohader, line 1: I'm not sure about the encoding here? Should it be plain instead?

stof: yes, this should be plain text, not HTML entities.

stof (approved):

PR #661, opened 2022-11-28
add info about vulns in phpxmlrpc/phpxmlrpc

gggeek:


phpxmlrpc/phpxmlrpc/2022-11-28-1.yaml
```diff
@@ -0,0 +1,8 @@
+title:     local file access in `Client:send` via manipulation of `$protocol` argument
+link:      https://github.com/gggeek/phpxmlrpc-extras/releases/tag/4.9.0
```
stof, line 2: This link is a 404.

Taking the composer package name and looking on Packagist, it looks like the right link should be https://github.com/gggeek/phpxmlrpc/releases/tag/4.9.0 (no `-extras` in the repo name).

Same for the other advisory.

stof (changes requested):

phpxmlrpc/phpxmlrpc/2022-11-28-1.yaml
gggeek: doh! copypasta

gggeek: Fixed in 2nd commit

PR #663, opened 2022-12-17
Added latest 3 TYPO3 extension security advisories

derhansen:
Added the following TYPO3 extension security advisories:

* https://typo3.org/security/advisory/typo3-ext-sa-2022-016
* https://typo3.org/security/advisory/typo3-ext-sa-2022-017
* https://typo3.org/security/advisory/typo3-ext-sa-2022-018

Note: The extension for CVE-2022-47408 does not use semantic versioning, so it is not possible to cover all affected versions.


fixpunkt/fp-newsletter/CVE-2022-47408.yaml
```diff
@@ -0,0 +1,14 @@
+title: 'TYPO3-EXT-SA-2022-017: Multiple vulnerabilities in extension "Newsletter subscriber management" (fp_newsletter)'
+link: https://typo3.org/security/advisory/typo3-ext-sa-2022-017
+cve: CVE-2022-47408
+branches:
+    master:
+        time: 2022-12-13 10:48:00
+        versions: ['>=3.0.0', '<3.2.6']
```
stof, line 7: this should probably be `>=2.2.0, <3.2.6` instead, to reflect the fact that 2.2+ is also affected (and hasn't been patched until 3.2.6)

stof: ok, looking at packagist, there is no releases in the 2.2.1-2.4.0 range, so why is the advisory talking about them ?

derhansen: I will update version constraints as suggested.

Version 2.4.0 seems to be a release, that has never been tagged and published to packagist. It has however been uploaded to TYPO3 extension repository https://extensions.typo3.org/extension/fp_newsletter as zip-package.

fixpunkt/fp-newsletter/CVE-2022-47408.yaml
```diff
@@ -4,10 +4,7 @@ cve: CVE-2022-47408
 branches:
     master:
         time: 2022-12-13 10:48:00
-        versions: ['>=3.0.0', '<3.2.6']
-    2.x:
-        time: 2022-12-13 10:38:00
-        versions: ['>=2.0.0', '<2.1.2']
```
stof: This change looks wrong to me, as it removes 2.1.2 as a non-vulnerable version.

derhansen: Sorry, removed it by mistake.

ohader (approved):
Looks good to me (sorry, I somehow forgot about this PR)

stof (approved):

PR #688, opened 2023-04-26
Add security advisories for Silverstripe CMS 4.13.0

GuySartorelli:
I can do these in separate PRs if you prefer, but I didn't see anything about that in the README so I assume this is okay.

### DIFF
--- a/silverstripe/framework/CVE-2023-22729.yaml
+++ b/silverstripe/framework/CVE-2023-22729.yaml
@@ -1,0 +1,8 @@
+title:     "CVE-2023-22729 - Open redirect vulnerability on CMSSecurity relogin screen"
+link:      https://www.silverstripe.org/download/security-releases/cve-2023-22729
+cve:       CVE-2023-22729
+branches:
+    4.12.x:
+        time:     2023-04-25 23:30:41
+        versions: ['>=4.0.0', '<4.12.5']
+reference: composer://silverstripe/framework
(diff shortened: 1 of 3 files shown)


silverstripe/framework/CVE-2023-22729.yaml
```diff
@@ -0,0 +1,11 @@
+title:     "CVE-2023-22729 - Open redirect vulnerability on CMSSecurity relogin screen"
+link:      https://www.silverstripe.org/download/security-releases/cve-2023-22729
+cve:       CVE-2023-22729
+branches:
+    4.0.x:
+        time:     2023-04-26 00:44:05
+        versions: ['>=4.0.0', '<4.12.5']
```
stof, line 7: Those entries for 4.0.x are useless. The range is already covered by you 4.12.x range (in all of them)

GuySartorelli: Fair enough, though in that case the readme should probably be updated to clarify what the "branches" information is for. My understanding from what is written is that each affected branch should have its own entry. 

Regardless, I'll update the pr to remove the 4.0.x branches tomorrow.

stof: We need multiple branches only when the issue gets patched in multiple branches and so a single range is not enough. Auditing tools will compare the version you use to those ranges.

GuySartorelli: Fair enough. I think the readme should be more explicit about that, but that's another discussion. I'll make the change to orrow as mentioned earlier.

stof: don't hesitate to open a PR improving the readme :)

GuySartorelli: I honestly don't feel like I understand the intention of the "branches" keys well enough to update the readme 😅 I'll leave that to someone who knows this project better.

GuySartorelli: `4.0.x` branches removed

PR #703, opened 2023-12-16
 Add CVE-2022-48614 for mediawiki/semantic-media-wiki 

jorgsowa:
Reference: https://nvd.nist.gov/vuln/detail/CVE-2022-48614

### DIFF
--- a/mediawiki/semantic-media-wiki/CVE-2022-48614.yaml
+++ b/mediawiki/semantic-media-wiki/CVE-2022-48614.yaml
@@ -1,0 +1,7 @@
+title: 'Cross-site Scripting in Semantic MediaWiki'
+link: 'https://nvd.nist.gov/vuln/detail/CVE-2022-48614'
+cve: CVE-2022-48614
+branches:
+    4.x:
+        versions: ['<4.0.2']
+reference: 'composer://mediawiki/semantic-media-wiki'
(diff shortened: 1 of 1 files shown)


mediawiki/semantic-media-wiki/CVE-2022-48614.yaml
```diff
@@ -0,0 +1,8 @@
+title: 'Cross-site Scripting in Semantic MediaWiki'
+link: 'https://nvd.nist.gov/vuln/detail/CVE-2022-48614'
```
xabbuh, line 2: I would link to https://www.semantic-mediawiki.org/wiki/Semantic_MediaWiki_4.0.2 here

xabbuh (changes requested):

mediawiki/semantic-media-wiki/CVE-2022-48614.yaml
stof: I agree

jorgsowa: There are already some links in the repository which are not working. I think more reliable are links to NVD.

Maybe the solution would be to allow for more links and/or references? What do you think?

stof (approved):

PR #740, opened 2025-01-28
Add CVE-2025-24856

derhansen:


causal/oidc/CVE-2025-24856.yaml
```diff
@@ -0,0 +1,8 @@
+title: 'TYPO3-EXT-SA-2025-001: Account Takeover in extension "OpenID Connect Authentication" (oidc)'
+link: https://typo3.org/security/advisory/typo3-ext-sa-2025-001
+cve: CVE-2025-24856
+branches:
+    main:
+        time: 2025-01-27 18:56:00
+        versions: ['3.0.0']
```
xabbuh, line 7:
```suggestion
        versions: ['>=3.0.0', '<4.0.0']
```

Like this?

naderman: There's literally only a single 3.0.0 release so kind of would have worked too ;-)

naderman:
After seeing this applied on packagist.org - @derhansen is it accurate that version 2.1.0 is safe to use? https://packagist.org/packages/causal/oidc#2.1.0

derhansen:
> After seeing this applied on packagist.org - @derhansen is it accurate that version 2.1.0 is safe to use? https://packagist.org/packages/causal/oidc#2.1.0

@naderman Thanks for the ping. According to the extension author, version 2.1.0 should be save. I however re-checked the code now and also found the vulnerability in version 2.1.0, so this version must also be marked as insecure. I discussed this with the extension author and will provide a follow up PR to update CVE-2025-24856 in this repository.

PR #743, opened 2025-01-30
Add unfixed vulnerabilities from Snyk

Rudloff:
validator.php requires an upper bound but these vulnerabilities are unfixed.


Rudloff:
Should I use the next major as upper bound to show that this is not fixed? Or an arbitrary high version number like 999?

PR #747, opened 2025-04-10
Silverstripe vulnerabilities April 2025

emteknetnz:

PR #748, opened 2025-04-10
Fix dates in silverstripe vulnerabilities

glaubinix:
Fixing commit dates in https://github.com/FriendsOfPHP/security-advisories/pull/747 where month and day was switched


glaubinix:
Thanks!

xabbuh:
Good catch, thanks @glaubinix!

PR #750, opened 2025-04-17
add CVE-2025-31481 and CVE-2025-31485 for API Platform

xabbuh:


api-platform/core/CVE-2025-31481.yaml
```diff
@@ -0,0 +1,14 @@
...
+branches:
+    3.4:
+        time:     2025-04-03 15:02:00
+        versions: ['>=3.4.0', '<3.4.17']
+    4.0:
+        time:     2025-04-03 15:02:00
+        versions: ['>=4.0.0', '<4.0.22']
+    4.1:
+        time:     2025-04-03 15:03:00
+        versions: ['>=4.1.0', '<4.1.5']
```
xabbuh, line 13: @soyuka Can you confirm that this is correct? The advisory does not talk about API Platform 4.1, but from what I see 4.1.5 is the first 4.1 release containing the patch.

xabbuh:
@soyuka Also, are all minor versions before 3.4 affected by these issues?

api-platform/core/CVE-2025-31481.yaml
```diff
@@ -0,0 +1,14 @@
+title:     "GraphQL query operations security can be bypassed"
+link:      https://github.com/advisories/GHSA-cg3c-245w-728m
+cve:       CVE-2025-31481
+branches:
+    '3.4':
+        time:     2025-04-03 15:02:00
+        versions: ['>=3.4.0', '<3.4.17']
```
stof, line 7: the GitHub advisory says `<3.4.17` is affected, so I would remove the lower bound here

soyuka:
Yes I could reproduce in 2.7 and I assume it's there since we introduced graphql.

api-platform/core/CVE-2025-31481.yaml
soyuka: Correct

stof: @soyuka you should probably update the GitHub advisory (both in the repository-level advisory and in their global advisory databases, as they are not automatically synchronized for updates to existing advisories)

xabbuh: Oh, and while you are at it, you should probably also adjust the affected 3.4/4.0 versions. Currently it states `<3.4.16` while that should rather be `<=`.

stof (approved):

PR #751, opened 2025-05-19
Add CVE-2025-47964 for Symfony UX Twig/Live Component

smnandre:
Related blog post: https://symfony.com/blog/symfony-ux-cve-2025-47946-unsanitized-html-attribute-injection-via-componentattributes

### DIFF
--- a/symfony/ux-live-component/CVE-2025-47946.yaml
+++ b/symfony/ux-live-component/CVE-2025-47946.yaml
@@ -3,6 +3,6 @@ link:   https://symfony.com/blog/symfony-ux-cve-2025-47946-unsanitized-html-attr
 cve:    CVE-2025-47946
 branches:
     2.x:
-        time:       2025-05-19 14:05:00
+        time:       2025-05-19 12:05:00
         versions:   ['<2.25.1']
 reference: composer://symfony/ux-live-component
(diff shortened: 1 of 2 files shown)


symfony/ux-live-component/CVE-2025-47946.yaml
```diff
@@ -0,0 +1,8 @@
+title:  "symfony/ux-live-component Unsanitized HTML attribute injection via ComponentAttributes"
+link:   https://symfony.com/blog/symfony-ux-cve-2025-47946-unsanitized-html-attribute-injection-via-componentattributes
+cve:    CVE-2025-47946
+branches:
+    2.x:
+        time:       2025-05-19 12:05:00
+        versions:   ['<2.25.1']
+reference: composer://symfony/ux-live-component
```
stof, line 8: Is ux-live-component actually impacted directly or only through its dependency on ux-twig-component (which would be reported anyway through the other advisory) ?

smnandre: I'd say directly as attribute were created and rendered directly from ux-live-component... no ?

stof: what happens if you use symfony/ux-live-component 2.24 with symfony/ux-twig-component 2.25.1 ? Would we still have the security issue, or would be it solved by the patch in symfony/ux-twig-component 2.25.1 ? That's the main question here (if updating only symfony/ux-twig-component to 2.25.1 removes the security vulnerability, we don't need an advisory for symfony/ux-live-component).

The fact that symfony/ux-live-component might be buggy (due to some double-escaping in places where it was working around the TwigComponent vulnerability before) when installing it alongside the patched ux-twig-component is a different topic (as this is not about a security issue)

smnandre: > ymfony/ux-live-component 2.24 with symfony/ux-twig-component 2.25.1

It would crash in multiple LiveComponent places, as ComponentAttributes would not have its second required argument when instancied (in live compoonent code)

stof (approved):

PR #754, opened 2025-11-12
add entries for CVE-2025-64500

xabbuh:


stof (approved):

carlinchisart:
Hi all, can we define the severity of this?  composer audit show null 
image

naderman:
This repository here does not currently track or determine severity in any way. In my opinion rightly so, as the typical severity ratings are usually pretty useless, as severity in the end depends too much on context of the specific use or integration with the vulnerable piece of software.

Composer/packagist.org do support severity values for other databases that supply them, but by default the severity is reported as null, which essentially means "undetermined".

PR #763, opened 2026-04-18
Add CVE-2026-24765

sebastianbergmann:


phpunit/phpunit/CVE-2026-24765.yaml
```diff
@@ -0,0 +1,20 @@
+title:     Unsafe Deserialization in PHPT Code Coverage Handling
+link:      https://github.com/sebastianbergmann/phpunit/security/advisories/GHSA-vvj3-c3rp-c85p
+cve:       CVE-2026-24765
+branches:
+    8.5.x:
+        time:     2026-01-27 05:21:14
+        versions: ['>=8.0.0', '<8.5.52']
```
glaubinix: @sebastianbergmann the GitHub advisory states that any version `<8.5.52` is affected by this. You are setting this to `>=8.0.0` saying that e.g. 7.x versions aren't affected by this. Just confirming that this is intentional.

stof (approved):

phpunit/phpunit/CVE-2026-24765.yaml
sebastianbergmann: I do not know whether PHPUnit <= 7 is affected or not and PHPUnit <= 7 do not get any changes/fixes anymore.

stof: The code in 7.5.20 looks to be the same than the code in 8.5 before the patch, so I think it is safe to assume it is also vulnerable. Removing this lowest bound to mark PHPUnit 7 and older as vulnerable makes sense IMO.

PR #762, opened 2026-04-18
Add GHSA-qrr6-mg7r-m243

sebastianbergmann:

PR #764, opened 2026-04-18
Add option to validate a single file

sebastianbergmann:
Validate a single YAML file

`validator.php` now accepts an optional path to a YAML file as its first argument. When provided, only that file is validated instead of the entire repository.

Usage:

```
php validator.php                                           # validate everything (unchanged)
php validator.php phpunit/phpunit/GHSA-qrr6-mg7r-m243.yaml  # validate single file
```

Relative and absolute paths are both accepted. If the given file does not exist, the command prints an error and exits with a non-zero status.

This makes it much faster to check a single advisory while authoring or reviewing it, without waiting for the full directory scan.

I found this helpful while working on https://github.com/FriendsOfPHP/security-advisories/pull/762.


glaubinix (approved):
Thank you! Looks helpful and works as advertised!

PR #769, opened 2026-05-20
Add advisories for Twig Security Release 2026-1

nicolas-grekas:
From https://symfony.com/blog/twig-3-26-0-released


jderusse (approved):

glaubinix (commented):
@nicolas-grekas all commit times say 2026-05-13 08:00:00 but looking at the Twig repo none of the fixes landed in 3.x before today 08:00:00. Think it's totally fine to not have exact commit times here as you all have enough work already but would at least set this to 2026-05-20 08:00:00 otherwise the time property becomes useless/harmful

CVE + titles seem to match commits. Links are still 404 though

nicolas-grekas:
Dates updated. The link will be fixed once we publish the corresponding blog posts. Should happen soon.

glaubinix (approved):
Thank you!

stof (approved):

PR #780, opened 2026-06-03
Add GHSA-2wwr-9x6f-88gp

javiereguiluz:


easycorp/easyadmin-bundle/GHSA-2wwr-9x6f-88gp.yaml
```diff
@@ -0,0 +1,10 @@
+title:     Path traversal and reflected XSS in Flag and Icon Twig components
+link:      https://github.com/EasyCorp/EasyAdminBundle/security/advisories/GHSA-2wwr-9x6f-88gp
+branches:
+    4.x:
+        time:     ~
+        versions: ['>=4.0.0', '<4.29.10']
```
glaubinix: The GitHub advisory states that all versions `<=4.29.9` are affected. Which one is true?

easycorp/easyadmin-bundle/GHSA-2wwr-9x6f-88gp.yaml
```diff
@@ -0,0 +1,10 @@
...
+    5.x:
+        time:     ~
```
glaubinix:
```suggestion
        time:  2026-05-28 18:30:00
```

easycorp/easyadmin-bundle/GHSA-2wwr-9x6f-88gp.yaml
```diff
@@ -0,0 +1,10 @@
+title:     Path traversal and reflected XSS in Flag and Icon Twig components
+link:      https://github.com/EasyCorp/EasyAdminBundle/security/advisories/GHSA-2wwr-9x6f-88gp
+branches:
+    4.x:
+        time:     ~
```
glaubinix:
```suggestion
        time:  2026-05-28 18:30:00
```

easycorp/easyadmin-bundle/GHSA-2wwr-9x6f-88gp.yaml
javiereguiluz: They should be equivalent ... but I updated this to use the same as in the security advisory.

javiereguiluz:
I did the requested changes. Thanks for reviewing this.

easycorp/easyadmin-bundle/GHSA-2wwr-9x6f-88gp.yaml
```diff
@@ -0,0 +1,10 @@
+title:     Path traversal and reflected XSS in Flag and Icon Twig components
+link:      https://github.com/EasyCorp/EasyAdminBundle/security/advisories/GHSA-2wwr-9x6f-88gp
+branches:
+    4.x:
+        time:     2026-05-28 18:30:00
+        versions: ['<=4.29.9', '<4.29.10']
```
stof: having 2 upper bounds does not make sense.

easycorp/easyadmin-bundle/GHSA-2wwr-9x6f-88gp.yaml
glaubinix: I think something went wrong with edit there. But this is not what i was asking. The GH security advisory states that versions 3.x, 2.x, 1.x are also affected by this. `['>=4.0.0', '<4.29.10']` says that only version 4.x (and 5.x) are affected.

easycorp/easyadmin-bundle/GHSA-2wwr-9x6f-88gp.yaml
```diff
@@ -0,0 +1,10 @@
...
+    5.x:
+        time:     2026-05-28 18:30:00
+        versions: ['<=5.0.9', '<5.0.10']
```
stof: this needs a lower bound `>=5.0.0`, otherwise the range would match the 4.29.10 version.

stof (changes requested):

easycorp/easyadmin-bundle/GHSA-2wwr-9x6f-88gp.yaml
javiereguiluz: I'm sorry. It's the first time I do this. I've updated both the GitHub advisory and this PR, so hopefully this is better now. Thanks.

easycorp/easyadmin-bundle/GHSA-2wwr-9x6f-88gp.yaml
javiereguiluz: Yes, I mixed things up. It should be better now. Thanks!

stof (approved):

stof:
Once you get a CVE id for this advisory (if you requested one), please submit another PR to update the advisory here with the CVE info.

glaubinix:
@stof now the constraint in the GitHub advisory is `>=4.0.0, <4.29.10` but here it is `['<4.29.10']`

PR #783, opened 2026-06-03
Add advisory for CVE-2026-48019 in laravel/framework

pavons-netsons:
https://github.com/laravel/framework/security/advisories/GHSA-5vg9-5847-vvmq


stof (changes requested):
As Laravel also ships subtree packages for each of their component, an advisory should also be added for the `illuminate/mail` package, so that `composer audit` can detect vulnerable versions when using the component package rather than the whole framework package.

stof (approved):

PR #785, opened 2026-06-04
Improve validator performance

javiereguiluz:
When I recently added a new advisory to this repo, I run the `validator.php` script to ensure it's correct.

However, I found that it was a bit slow to run (~24 seconds). I see that the validator is making a network call to search on Packagist per advisory, instead of per package.

1,166 files in this repo trigger a network search, but they only represent 311 distinct packages, so we can avoid 73% of network calls with some memoization.

Performance comparation:

```bash
$ time php validator.php

Before:  23.92s user 3.69s system 98% cpu 28.027 total
After:    6.54s user 0.98s system 88% cpu  8.468 total
```

### DIFF
--- a/validator.php
+++ b/validator.php
@@ -31,6 +31,7 @@ final class Validate extends Command
 {
     private Parser $parser;
     private array $composerRepositories = array();
+    private array $composerPackageExists = array();
     private Config $composerConfig;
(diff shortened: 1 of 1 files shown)


glaubinix (approved):
Amazing, thank you!

PR #789, opened 2026-06-17
Adding GHSA-3prj-6hqw-cm82, GHSA-5739-39v2-5754, GHSA-6vvh-pxr4-25r7 …

hostep:
…& GHSA-jc38-x7x8-2xc8 - web-token/jwt-framework

First time contributor here, let me know if I did something wrong.
Not sure about branches format. Anyway, if this needs some tweaks, let me know.

Adding these files since `composer audit` at the moment isn't outputting these security vulnerabilities.

### DIFF
--- a/web-token/jwt-framework/GHSA-3prj-6hqw-cm82.yaml
+++ b/web-token/jwt-framework/GHSA-3prj-6hqw-cm82.yaml
@@ -1,0 +1,14 @@
+title: 'PBES2-HS*+A*KW unwrap accepts an unbounded p2c iteration count, enabling CPU-amplification denial of service'
+link: 'https://github.com/web-token/jwt-framework/security/advisories/GHSA-3prj-6hqw-cm82'
+cve: ~
+branches:
+    3.x:
+        versions: ['<=3.4.9']
+        time: 2026-06-06 16:26:43
+    4.0.x:
+        versions: ['>=4.0.0', '<=4.0.6']
+        time: 2026-06-06 16:26:43
+    4.1.x:
+        versions: ['>=4.1.0', '<=4.1.6']
+        time: 2026-06-06 16:26:43
+reference: 'composer://web-token/jwt-framework'
--- a/web-token/jwt-framework/GHSA-5739-39v2-5754.yaml
+++ b/web-token/jwt-framework/GHSA-5739-39v2-5754.yaml
@@ -1,0 +1,14 @@
+title: 'RSA1_5 (RSAES-PKCS1-v1_5) decryption lacks implicit rejection, exposing a Bleichenbacher/Marvin padding oracle'
+link: 'https://github.com/web-token/jwt-framework/security/advisories/GHSA-5739-39v2-5754'
+cve: ~
+branches:
+    3.x:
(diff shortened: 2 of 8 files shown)


web-token/jwt-framework/GHSA-3prj-6hqw-cm82.yaml
```diff
@@ -0,0 +1,14 @@
+title: 'PBES2-HS*+A*KW unwrap accepts an unbounded p2c iteration count, enabling CPU-amplification denial of service'
+link: 'https://github.com/web-token/jwt-framework/security/advisories/GHSA-3prj-6hqw-cm82'
+cve: ~
+branches:
+    3.x:
```
stof: I would name it `3.4.x`, to match the name of the dev branch in which the fix was merged.

web-token/jwt-framework/GHSA-3prj-6hqw-cm82.yaml
```diff
@@ -0,0 +1,14 @@
...
+        versions: ['<=3.4.9']
+        time: 2026-06-06 16:26:43
+    4.0.x:
+        versions: ['>=4.0.0', '<=4.0.6']
+        time: 2026-06-06 16:26:43
+    4.1.x:
+        versions: ['>=4.1.0', '<=4.1.6']
+        time: 2026-06-06 16:26:43
+reference: 'composer://web-token/jwt-framework'
```
stof, line 14: As `web-token/jwt-framework` is a composer package that is published for the monorepo (replacing subpackages), we need similar advisories for the `web-token/jwt-library` package (which is consistent with the GitHub advisory marking both package names as impacted)

web-token/jwt-framework/GHSA-5739-39v2-5754.yaml
```diff
@@ -0,0 +1,14 @@
...
+cve: ~
+branches:
+    3.x:
+        versions: ['<=3.4.9']
+        time: 2026-06-06 16:27:24
+    4.0.x:
+        versions: ['>=4.0.0', '<=4.0.6']
+        time: 2026-06-06 16:27:24
+    4.1.x:
+        versions: ['>=4.1.0', '<=4.1.6']
```
stof: We generally define ranges with an excluside upper bound, by using the version releasing the patch as upper bound. It makes it easier to see when the patch was released when reading the advisory.

hostep:
Thanks @stof, that's very useful feedback!

web-token/jwt-framework/GHSA-5739-39v2-5754.yaml
naderman: @stof any chance you can simply add this to the validator in this repo? this keeps coming up.

naderman (approved):

web-token/jwt-framework/GHSA-5739-39v2-5754.yaml
stof: this seems totally doable.

stof: See #790

PR #790, opened 2026-06-19
Enforce using the patched version as upper bound

stof:
This makes it easier to see which version is the patched version as the constraint references it directly.
This rule applies only to patched vulnerabilities.

Putting this in the validator avoids bringing this during human reviews.
The PR fixes old advisories where we were not yet asking for this convention (none of the advisories added after 2021 were affected, which is a sign that we enforce it during review already).


glaubinix (approved):
Thank you!

naderman:
@stof so these in principle look fine. How much effort do we want to put into verifying these all actually use a 3 digit versioning scheme? Meaning if the old constraint was <= 1.2.3 and you changed it to <1.2.4, do we want to make sure there wasn't a 1.2.3.1 release also fixing the issue, or do we just assume this is fine?

stof:
@naderman for all packages I modified, I looked at the advisory URL to see the patched version or at packagist.org to see what was the next release (when the advisory link is now dead).
The only one where I'm unsure is `cart2quote/module-quotation`, which is not a package hosted on packagist.org and where the advisory link is also dead.

naderman:
@stof Great, thank you!

PR #797, opened 2026-06-24
Fix the CI validation

stof:
The `matyhtf/framework` package has been deleted from Packagist.

### DIFF
--- a/matyhtf/framework/CVE-2021-43676.yaml
+++ b/matyhtf/framework/CVE-2021-43676.yaml
@@ -2,6 +2,7 @@ title: Path manipulation
 link: https://github.com/advisories/GHSA-mh9j-v6mq-pfch
 cve: CVE-2021-43676
 reference: composer://matyhtf/framework
+composer-repository: false # package has been deleted: https://packagist.org/transparency-log?actor=&user=&vendor=&package=matyhtf%2Fframework&datetime_from=&datetime_to=
 branches:
     '3.0':
         time: 2022-03-17 16:15:10


matyhtf/framework/CVE-2021-43676.yaml
```diff
@@ -2,6 +2,7 @@ title: Path manipulation
 link: https://github.com/advisories/GHSA-mh9j-v6mq-pfch
 cve: CVE-2021-43676
 reference: composer://matyhtf/framework
+composer-repository: false # package has been deleted: https://packagist.org/transparency-log?actor=&user=&vendor=&package=matyhtf%2Fframework&datetime_from=&datetime_to=
```
glaubinix, line 5: Thanks for adding the link to the transparency log! 

One note the log is currently still quite slow when using the package search. If you know the vendor name then adding that one to the search query uses an index which make it fast

stof: oh, I missed the fact that https://github.com/composer/packagist/pull/1768 was not merged yet.

Is it worth doing another PR here to update the comment, or can I consider that this PR will be merged in Packagist (applying the optimization automatically) ? I don't think this comment will generate lots of traffic anyway.

glaubinix: The PR will definitely be merged in some form. Think still figuring out what else we can improve.

Wouldn't update the comment anymore. This was more for you when using the transparency log that you don't have to wait there for seconds/minutes until you see results.

stof: maybe you can merge it and do more optimizations in a separate PR (once you identify what is still slow *after* those changes).

PR #800, opened 2026-07-27
Add advisory for PHPCSUtils CVE-2026-65954

jrfnl:


jrfnl:
Sorry, but your validation seems wrong. `stable` is the branch name for the release branch of PHPCSUtils....

glaubinix:
Generally unsure about adding `stable` here as we'd then have to accept other branch names too just because someone calls their release branch `trunk`, `production`, or `releases`. Think the alternative here would be just to call it `1.x`. 

I am not sure in how far any tooling using this repo relies on accurate branch names though branch names can easily be deleted and/or recreated under a different name and become outdated once someone tags 2.0 and decides there is a need for two release branches and stable is 2.x only and a new 1.x covers older releases. @stof you were the last one to widen the regex. Any input here/

stof:
I think the branch name was used in https://github.com/fabpot/local-php-security-checker to handle the case of projects using a dev version.
I don't think `composer audit` cares about the branch name defined here. IIRC, packagist.org only cares about the version ranges when importing the advisories.

glaubinix:
Right, both Packagist/Composer and the roave/security-advisories package don't care about branches. Any preferences of expanding the regex with more branch names or falling back to 1.x in case packages uses a different branch name?

stof:
I suggest removing this regex validating the branch name entirely.
There is no need to implement any kind of fallback. The validator does not care either about the structure of the branch name outside of this validation rule.

stof:
See https://github.com/FriendsOfPHP/security-advisories/pull/803

jrfnl:
I will rebase and drop that extra commit after #803 has been merged.

glaubinix:
Thank you @stof! This PR is now ready to be rebased.

jrfnl:
Rebased & force pushed. Ready for merge again.

jrfnl:
Build failure is likely due to GH Actions being down: https://www.githubstatus.com/incidents/qcvjkzcs7j74

jrfnl:
Force pushed without changes to get a passing build. 🤞🏻

glaubinix (approved):
thank you!

PR #802, opened 2026-08-06
Add advisory for PHP_CodeSniffer CVE-2026-67434

jrfnl:
Open question: should old branches, like the `1.5` and `2.9` branches, still be mentioned here ?

### DIFF
--- a/squizlabs/php_codesniffer/CVE-2026-67434.yaml
+++ b/squizlabs/php_codesniffer/CVE-2026-67434.yaml
@@ -1,0 +1,11 @@
+title:     OS Command injection
+link:      https://github.com/PHPCSStandards/PHP_CodeSniffer/security/advisories/GHSA-hmqg-cxww-wqhq
+cve:       CVE-2026-67434
+branches:
+    3.x:
+        time:     2026-08-05 23:53:11
+        versions: ['<3.13.6']
+    4.x:
+        time:     2026-08-05 23:53:11
+        versions: ['>=4.0.0', '<4.0.2']
+reference: composer://squizlabs/php_codesniffer


glaubinix:
I take it all those branches are EOL/unmaintained? If that's the case then dropping them should be fine, so 
```
branches:
    3.x:
        time:     2026-08-05 23:53:11
        versions: ['<3.13.6']
    4.x:
        time:     2026-08-05 23:53:11
        versions: ['>=4.0.0', '<4.0.2']
```

If you want to keep the entries then the bellow would be better
```
branches:
    1.x:  #or 1.5.x
        time:     ~
        versions: ['<2.0.0']
    2.x:  # or 2.9.x
        time:     ~
        versions: ['>=2.0.0', '<3.0.0']
    3.x: 
        time:     2026-08-05 23:53:11
        versions: ['>=3.0.0', '<3.13.6']
    4.x:
        time:     2026-08-05 23:53:11
        versions: ['>=4.0.0', '<4.0.2']
```

jrfnl:
> I take it all those branches are EOL/unmaintained?

Yes, 1.x and 2.x are very much end of life (and have been for > 8 years).

> If that's the case then dropping them should be fine, so

Will do.

> If you want to keep the entries then the bellow would be better
> 
> ```
> branches:
>     1.x:  #or 1.5.x
>         time:     ~
>         versions: ['<2.0.0']
>     2.x:  # or 2.9.x
>         time:     ~
>         versions: ['>=2.0.0', '<3.0.0']
> ```

Noted, except that would result in non-existent branch names from being mentioned, which I believe is flagged by the validator. I know the `1..5` and `2.9` branch names are unconventional, but that's what we're stuck with and I'm not about to rename those unmaintained/historic branches.

stof:
Both packagist.org and `roave/security-advisories` (which are the 2 known consumers of this database nowadays as https://github.com/fabpot/local-php-security-checker is archived) only care about the (combined) version ranges. So you don't need to duplicate them.

stof (approved):

PR #817, opened 2026-09-01
Add a missing CVE ID

javiereguiluz:
GitHub folks just issued the CVE ID for this security issue, so let's add it. Thanks.


glaubinix (approved):
Thank you!

PR #820, opened 2026-09-09
Correct the two `react/http` advisories for `v1.11.1` to their correct version targetting

WyriHaximus:
Forgot to add the `=` in `<=1.11.0`.


glaubinix (commented):
Thanks!

